PR #15, opened 2022-03-14
Add `load-balancer_admin` role to `admin` user

nicolasbock:
Starting with Wallaby, Octavia insists on a system scoped token for
the admin user [1]. Instead, this change adds the `load-balancer_admin`
role to the `admin` user.

[1] https://docs.openstack.org/octavia/wallaby/configuration/policy.html


nicolasbock:
@dosaboy This role was added to Octavia in the Pike release and should be safe to add unconditionally.

PR #17, opened 2022-03-17
Add SSL configuration for placement API overlay

dnegreira:
Currently no SSL gets configured on placement API which causes issues with SSL enabled deployments, this small patch follows the similar configuration of keystone et all that configure SSL properly.

Signed-off-by: David Negreira <david.negreira@canonical.com>


dnegreira:
Found some issues with deployment, will see what is going on and will re-open if needed

PR #25, opened 2022-04-05
Add mongo access script

casual-lemon:
Tested and I was able to access mongodb

### DIFF
--- a/tools/mongo-access.sh
+++ b/tools/mongo-access.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+  
+machine=${1:-0}
+model=${2:-controller}
+
+read -d '' -r cmds <<'EOF'
+conf=/var/lib/juju/agents/machine-*/agent.conf
+user=`sudo grep tag $conf | cut -d' ' -f2`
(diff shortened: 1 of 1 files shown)


nicolasbock (commented):
It would be nice if the script would figure out what the primary controller machine is for HA setups, i.e. without specifying the machine the script would connect to the primary and when the machine is specified it connects to that particular machine.

nicolasbock (approved):

tools/mongo-access.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+  
+machine=${1:-0}
+model=${2:-controller}
+
```
sombrafam, line 5: Would be nice to have some basic parameter handling and help, just so when people run the first time don't need to open the script to understand the parameters.

sombrafam (changes requested):

tools/mongo-access.sh
nicolasbock: I agree, but I would also be ok with doing that in a follow-up PR.

PR #27, opened 2022-04-13
Fix keystone-saml setup tool

dosaboy:
This tool was using role "Member" which does not
exist in our current default setup and should actually
use "member".


nicolasbock (approved):
Thanks!

PR #28, opened 2022-04-15
Check for keystone-saml-mellon application

nicolasbock:
Add a check to the script to verify that the keystone-saml-mellon
application is deployed.


hemanthnakkina (approved):
LGTM!

Issue #30, opened 2022-04-20
Cannot deploy application "grafana": unknown option "admin_password"


marino-mrc:
Hi, I'm trying to deploy Openstack adding the lma overlay. Here is the output of the command:

`
~/stsstack-bundles/openstack$ ./generate-bundle.sh  --lma -s focal -r ussuri -n openstack --run
Creating Juju model openstack
Added 'openstack' model on stsstack/stsstack with credential 'marino-mrc' for user 'admin'

Created focal-ussuri bundle and overlays:
  + glance.yaml
  + keystone.yaml
  + neutron-gateway.yaml
  + neutron-openvswitch.yaml
  + neutron-ml2dns.yaml
  + neutron-ml2dns-gateway.yaml
  + neutron-ml2dns-openvswitch.yaml
  + mysql-innodb-cluster.yaml
  + mysql-innodb-cluster-router.yaml
  + placement.yaml
  + graylog.yaml
  + grafana.yaml
  + prometheus-openstack.yaml
  + prometheus-openstack-neutron-gateway.yaml

Command to deploy:
juju deploy ./b/openstack/openstack.yaml --overlay ./b/openstack/o/glance.yaml --overlay ./b/openstack/o/keystone.yaml --overlay ./b/openstack/o/neutron-gateway.yaml --overlay ./b/openstack/o/neutron-openvswitch.yaml --overlay ./b/openstack/o/neutron-ml2dns.yaml --overlay ./b/openstack/o/neutron-ml2dns-gateway.yaml --overlay ./b/openstack/o/neutron-ml2dns-openvswitch.yaml --overlay ./b/openstack/o/mysql-innodb-cluster.yaml --overlay ./b/openstack/o/mysql-innodb-cluster-router.yaml --overlay ./b/openstack/o/placement.yaml --overlay ./b/openstack/o/graylog.yaml --overlay ./b/openstack/o/grafana.yaml --overlay ./b/openstack/o/prometheus-openstack.yaml --overlay ./b/openstack/o/prometheus-openstack-neutron-gateway.yaml 
 
Located charm "cinder" in charm-hub, channel ussuri/edge
Located charm "mysql-router" in charm-hub, channel 8.0.19/edge
Located charm "elasticsearch" in charm-hub, channel stable
Located charm "filebeat" in charm-hub, channel stable
Located charm "glance" in charm-hub, channel ussuri/edge
Located charm "mysql-router" in charm-hub, channel 8.0.19/edge
[...]

PR #31, opened 2022-04-20
Skip prefix for non-osm grafana

dosaboy:
The osm and lma charms each have their own version of
the grafana charm where the lma one has no prefix in
the charmhub.

Resolves: #30

### DIFF
--- a/common/render.d/all
+++ b/common/render.d/all
@@ -16,7 +16,7 @@ render_mod_params ()
 
     local charms
     # get list of charms in this template
-    readarray -t charms<<<`sed -rn 's,.+ charm:\s+.+__CHARM_CH_PREFIX__(.+)\s*$,\1,p' $1`
+    readarray -t charms<<<`sed -rn 's,.+ charm:\s+.+__CHARM_CH_PREFIX__(__SKIP_PREFIX__)?(.+)\s*$,\2,p' $template`
 
     config_renderer=`mktemp`
     export INTERNAL_BUNDLE_CONFIG=`mktemp`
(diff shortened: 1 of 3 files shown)


common/render.d/all
```diff
@@ -28,8 +28,12 @@ render_mod_params ()
             channel=${CHARM_CHANNEL[$c]:-$service_default_channel}
             echo "-e '/__CHARM_CS_NS____CHARM_CH_PREFIX__${c}\s*$/a\ \ \ \ channel: $channel'" >> $config_renderer
 
-            prefix=${CH_PREFIXED_CHARMS[$c]:-""}
-            echo "-e 's,__CHARM_CS_NS____CHARM_CH_PREFIX__${c}\s*$,$prefix${c},'" >> $config_renderer
+            if `grep -q "__SKIP_PREFIX__$c" $template`; then
+                echo "-e 's,__CHARM_CS_NS____CHARM_CH_PREFIX____SKIP_PREFIX__${c}\s*$,${c},'" >> $config_renderer
```
nicolasbock, line 32: I am not familiar with if and when we might want to use the prefixed version of the `grafana` charm. I don't see how we would in case we wanted to.

dosaboy: As far as I know the prefixed is only ever for use with osm.

hemanthnakkina: Ed, 

How about LMA for Charmed Kubernetes. Is operator based grafana (same as OSM one) used for Charmed kubernetes lma stack?

dosaboy: @hemanthnakkina my understanding is that the osm prefixed charms are only for use with osm (they are developed and maintained separately to the others).

nicolasbock: Thanks for the update @dosaboy . Looks good to me then :)

nicolasbock (approved):

PR #38, opened 2022-06-20
Update ch channel versions for focal/impish/jammy

dosaboy:


nicolasbock (approved):
I didn't have to use a local copy, the `--force` option was enough.

(in reference to common/ch_channel_map/impish:9-10, I can't leave an inline comment for some reason)

PR #44, opened 2022-06-23
Use `ACTIVE_STANDBY` topology for > one compute

nicolasbock:
If more than one `nova-compute` unit is requested default to `ACTIVE_STANDBY`
for load balancers. This setting aligns more closely to what customers typically
use.


openstack/pipeline/02configure
```diff
@@ -68,6 +68,7 @@ do
     case "$1" in
         --num-compute)  #__OPT__type:<int>
             MOD_PARAMS[__NUM_COMPUTE_UNITS__]=$2
+            MOD_PARAMS[__LOADBALANCER_TOPOLOGY__]=ACTIVE_STANDBY
```
dosaboy, line 71: should you check that there are > 1?

nicolasbock: oops, yes, that would be helpful :/

nicolasbock: done

dosaboy (approved):

PR #45, opened 2022-06-25
Add script to look at logs from multiple units

nicolasbock:
The `juju-lnav` script is a helper script to call `lnav` (needs to be installed
via package or snap) on multiple machines, applications, or units. For example,

    $ ./tools/juju-lnav octavia:/var/log/octavia/*.log{,1}

Will open the current and the first rotated log files from all of the `octavia`
units.

Likewise

    $ ./tools/juju-lnav octavia:/var/log/octavia/*.log \
        nova-cloud-controller:/var/log/nova/*.log

Will load the logs on all `octavia` and all `nova-cloud-controller` units.

For more information on how to use `lnav` please visit lnav.org.

### DIFF
--- a/openstack/tools/juju-lnav
+++ b/openstack/tools/juju-lnav
@@ -1,0 +1,1 @@
+../../tools/juju-lnav
--- a/tools/juju-lnav
+++ b/tools/juju-lnav
@@ -1,0 +1,139 @@
+#!/bin/bash
+
+set -e -u
+
+declare -A MACHINES=()
+declare -A CONTAINERS=()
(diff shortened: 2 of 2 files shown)


dosaboy:
would this perhaps not be best published as a snap/deb so that people can use it without stsstack-bundles?

nicolasbock:
Yes that would be even more useful. Thanks for the suggestion @dosaboy !

nicolasbock:
I moved the script into its own repository

https://github.com/nicolasbock/juju-lnav

PR #46, opened 2022-06-25
Add bashate linting to CI

nicolasbock:


sombrafam (changes requested):
There are some unrelated changes on the patch. Would be nice to get those in another PR.

sombrafam:
Now I see that the changes are to make bashate pass. We are good then.

Issue #54, opened 2022-08-19
./configure generates wrong channel for barbican-vault charm


dnegreira:
Seems that the channel selection for `barbica-vault` is not being set correctly.

```
ERROR cannot deploy bundle: cannot resolve charm or bundle "barbican-vault": selecting releases: charm or bundle not found for channel "ussuri/edge", platform "amd64"
```
Current channels available are:

```
latest/stable
latest/edge 
yoga/stable 
xena/stable
```

PR #55, opened 2022-08-19
Fix channel for barbican-vault

dnegreira:
This change aims to add the correct channel for barbican-vault

Closes: #54

Signed-off-by: David Negreira <david.negreira@canonical.com>

### DIFF
--- a/common/ch_channel_map/bionic
+++ b/common/ch_channel_map/bionic
@@ -12,3 +12,7 @@ CHARM_CHANNEL[hacluster]=2.0.3/edge
 CHARM_CHANNEL[ovn-central]=20.03/edge
 CHARM_CHANNEL[ovn-chassis]=20.03/edge
 CHARM_CHANNEL[ovn-dedicated-chassis]=20.03/edge
+
+# At the moment of this patch, the barbican-vault only has {yoga,xena}/stable channels
+
+CHARM_CHANNEL[barbican-vault]=latest/stable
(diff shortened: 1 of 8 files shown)


common/ch_channel_map/bionic
```diff
@@ -4,6 +4,7 @@
 # Versions are based on https://docs.openstack.org/charm-guide/latest/project/charm-delivery.html
 
 CHARM_CHANNEL[vault]=1.6/edge
+CHARM_CHANNEL[barbican-vault]=latest/edge
```
nicolasbock, line 7: Did you find documentation on what channel we should be using here?

dnegreira: Unfortunately, I have not found any documentation that points to any specific mentions of channel for barbican-vault should be used for which OpenStack version / Ubuntu series other than it's only supported on Openstack Rocky and above.

FWIW, the [barbican-vaultcharm](https://charmhub.io/barbican-vault) only has two related OpenStack version channels there: 
- `yoga-stable`
- `xena/stable`

Should I adjust those on this PR?

The only other two channels are `latest/stable` and `latest/edge`.

nicolasbock: Let's see what the product team thinks. Could you ask them and find out what channel we should be using for what version of OS / OpenStack?

dnegreira: @nicolasbock I've force pushed a new commit after checking with the OS team what are the right channel versions to use.

nicolasbock (approved):
Thank you @dnegreira

PR #65, opened 2022-09-28
Add option to specify IP for `mongo-access.sh`

nicolasbock:
This change adds an option to `mongo-access.sh` so that an IP address can be specified. This allows the use of this script even in cases when the `juju` controller is not responding.


tools/mongo-access.sh
```diff
@@ -29,6 +34,10 @@ EOF
             shift
             model="$1"
             ;;
+        --IP)
```
dosaboy, line 37: does it need to be uppercase?

nicolasbock: No, I made it lower case.

casual-lemon (dismissed):
LGTM +1

tools/mongo-access.sh
```diff
@@ -4,6 +4,8 @@ set -u
 
 : "${machine:=0}"
 : "${model:=controller}"
+: "${IP_address:=}"
```
sombrafam, line 7: ip_address

tools/mongo-access.sh
```diff
@@ -29,6 +34,10 @@ EOF
             shift
             model="$1"
             ;;
+        --ip)
+            shift
+            IP_address="$1"
```
sombrafam, line 39: same

sombrafam (changes requested):
Hey Nicolas, just a small change on the var name

tools/mongo-access.sh
nicolasbock: Done

sombrafam (approved):
LGTM

sombrafam (approved):

PR #70, opened 2022-10-10
Explcitily specify source=distro instead of blank

lathiat:
Charmhub charms (including git master) now set the charm default
openstack-origin/source option to the matching openstack release name
(e.g. 'wallaby') instead of the previous default of 'distro'.

As a result, if you try to test a git master charm against an older
OpenStack release that uses distro (e.g. ussuri) you will accidentally
upgrade that charm from the current release (e.g. focal-ussuri) to the
latest release in git master (e.g. focal-wallaby).

While this is not technically supported and some charms are now removing
support for older releases from git master, it's still OK to test some
charms like this and even if not supported it usually incorrectly
upgrades that application which breaks the environment and is difficult
to rollback leaving you having to spend another 30-60+ minutes
re-deploying your test environment.

To prevent that happening by default, explicitly specify 'distro'
instead of leaving source blank.

### DIFF
--- a/common/helpers
+++ b/common/helpers
@@ -761,7 +761,7 @@ if [ -n "$ppa" ]; then
 elif ! ltsmatch "$series" "$release" && ! nonltsmatch "$series" "$release" ; then
     source="cloud:${series}-${release}"
 else
-    source=""
+    source="distro"
 fi
 if [ -n "$pocket" ]; then
     if [ -n "$source" ]; then


dosaboy:
I'd like to make sure I'm clear why you would want to test a git master charm against an older release of Openstack. The way the charms are published is that they only support (and are tested for use with) the release of Openstack corresponding to the channel they are released in. I assume your use case is where a patch lands in master that you might want to backport to an earlier release but you want to test it with that release first. In that case mightn't it be better to just to a local backport onto the corresponding stable branch and deploy/upgrade that?

lathiat:
As you say the intent is that you shouldn't use a master charm against an older version, and they are just now starting to drop support for older versions in the most recent git masters, but that's still not true of all charms they kept all the code as is supporting many versions and so for initial testing it was a good shortcut to just deploy git master against a reproducer that you had.

But I'd be fine if that didn't work and just failed - I don't really want that specific use case, the problem I had is, that workflow (which used to work) if you forget and deploy the wrong charm version (which is super easy to do when deploying from source as there is no check that you requested the right 'channel') into a focal-ussuri cloud it automaticlaly starts and finishes an upgrade whether or not that was a sensible upgrade and whether or not the other dependant components are up to date.

Once this happens its a very expensive mistake as you need to tear down the environment and rebuild it again. At best that takes 30+ minutes for a re-deploy and sometimes much longer if the steps to reproduce the issue are long. This happens unexpectedly because of the default openstack-origin change.

Ideally, to me, until such a time as they actually remove that option - I'd prefer it only changed when I expected it to. That happens already when you use any other version than usurri (or other version that sets openstack-origin to distro). If you deploy victoria onwards and you make the same mistake the same upgrade doesn't happen because openstack-origin is expliclyt set in the bundle (the charm just might fail to work properly, but in most cases force upgrading to the correct charm will have you working again) but specifically only against usurri you end up with this accidental unintended upgrade that catches you out.

So by making this change, we ensure that the openstack-origin only changes when you explicitly change it intentionally and can save people time from accidental mistakes.

So yes you're right - in theory you shouldn't do it - but the mistake is so easy to make (I made the same mistake 3 times, once even right after already having done it before) for no good reason and it's a weird behaviour inconsistency between deploying usurri or any other later version. So making it explicit will hopefully save others from the same painful mistake.

At some point they'll remove the option all together and it won't help there but it's not going to hurt either. I did consider whether it would make it 'harder' to have switching charm channel only do the upgrade - which might be true except, it's only true from usurri. Deploying any other version of openstack leaves openstack-origin explicitly set and not using the charm default so by 'not' doing this we have that same inconsistency. It would be nice if it was explicitly consistent and didn't only sometimes rely on the default being what we thought it was.

nicolasbock:
@lathiat What "downside" would that change have if any? It doesn't look to me
like it would change current behavior. It would simply spell out what one
would normally get anyway, right?

lathiat:
Yes I don't see any downside, it's doing the same you normally get anyway. It's just making it explicit so we don't have it change on us by accident when upgrading charms particularly to the wrong version.

nicolasbock:
Hi @lathiat. Do we still need this change?

lathiat:
Since #120 switched focal-ussuri to use OVN 22.03, it now deploys ovn-central with source=blank, which uses the default "yoga" and breaks the deployment. It needs to be set explicitly to distro (e.g. from https://docs.openstack.org/charm-guide/2023.1/project/procedures/ovn-upgrade-2203.html)

So this would also fix that problem, in addition to the problem I had previously mentioned.


It seems newer charms support setting the openstack release, e.g. 'yoga', regardless of which ubuntu base version you are on. So I guess we could also adopt that for maybe yoga+ but to keep supporting deploying older releases (a specific goal of stsstack-bundles) we're better off to use 'distro' for ussuri and older I think.. I'm not sure if all the charms supported a plain 'ussuri' in ussuri. I guess we could test.

PR #73, opened 2022-10-18
Add script to look at logs from multiple units

nicolasbock:
The `juju-lnav` script is a helper script to call `lnav` (needs to be installed
via package or snap) on multiple machines, applications, or units. For example,

    $ ./tools/juju-lnav octavia:/var/log/octavia/*.log{,1}

Will open the current and the first rotated log files from all of the `octavia`
units.

Likewise

    $ ./tools/juju-lnav octavia:/var/log/octavia/*.log \
        nova-cloud-controller:/var/log/nova/*.log

Will load the logs on all `octavia` and all `nova-cloud-controller` units.

For more information on how to use `lnav` please visit lnav.org.

### DIFF
--- a/tools/juju-lnav
+++ b/tools/juju-lnav
@@ -1,0 +1,139 @@
+#!/bin/bash
+
+set -e -u
+
+declare -A MACHINES=()
+declare -A CONTAINERS=()
+declare -A UNITS=()
+
+install_ssh_key() {
+    local unit
+    for unit in $@; do
+        echo "installing ssh key for ${unit}"
+        cat ~/.ssh/id_rsa.pub | timeout 10 juju ssh ${unit} -- sudo tee --append /root/.ssh/authorized_keys
+    done
+}
+
+get_application_units() {
+    local application=$1
+    local -a units=()
+    local i
+    for i in "${!UNITS[@]}"; do
+        if [[ ${i} =~ ^${application}/[0-9]+$ ]]; then
+            units=( "${units[@]}" "${i}" )
+        fi
+    done
+    echo "${units[@]}"
+}
+
+expand_unit() {
+    local -a split_string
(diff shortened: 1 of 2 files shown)


tools/juju-lnav
```diff
@@ -0,0 +1,139 @@
...
+fi
+
+while (( $# > 0 )); do
+    case $1 in
+        -h|--help)
+            cat <<EOF
+Usage:
+
+$(basename "$0") UNIT:LOGFILES [UNIT:LOGFILES [UNIT:LOGFILES [...]]]
+
```
sombrafam, line 117: can you add a working example here? like the one in the commit?

tools/juju-lnav
```diff
@@ -0,0 +1,139 @@
...
+Options:
+
+UNIT        is the name of the unit
+LOGFILES    is a regular shell GLOB
```
sombrafam, line 121: 'is any text file in the juju init'?

tools/juju-lnav
```diff
@@ -0,0 +1,139 @@
...
+}
+
+declare -a lnav_arguments=()
+juju_info_loaded=0
+
+if ! command -v lnav > /dev/null; then
+    cat <<EOF
+Please install lnav with
+
+sudo snap install --edge lnav
```
sombrafam, line 103: also run: sudo snap connect lnav:ssh-keys

nicolasbock: Fixed

tools/juju-lnav
nicolasbock: Yes pretty much although lnav as parsers for particular log file types. But yes, it can read any text file.

tools/juju-lnav
nicolasbock: Done

PR #76, opened 2022-10-19
Adds support for cinder-lvm

sombrafam:


overlays/cinder-lvm.yaml
```diff
@@ -0,0 +1,9 @@
+applications:
+  cinder-lvm:
+    charm: __CHARM_STORE____CHARM_CS_NS____CHARM_CH_PREFIX__cinder-lvm
+    options:
+      alias: stsstack
+      allocation-type: thin
+      block-device: /var/lib/lvm_pool0.img|20G
```
nicolasbock, line 7: The default flavor is `m1.small` which has a 20G disk in total. Either increase the flavor used or make this smaller.

nicolasbock (changes requested):

overlays/cinder-lvm.yaml
sombrafam: Good catch. This did not come to the surface because the volumes are thin. I made that smaller.

nicolasbock (approved):
Thanks @sombrafam !

nicolasbock (approved):

PR #81, opened 2022-11-09
Add script to create volumes needed by kafka units

drencrom:
This script creates two volumes for each kafka unit and attaches them to the units.

### DIFF
--- a/kafka/tools/create_volumes.sh
+++ b/kafka/tools/create_volumes.sh
@@ -1,0 +1,17 @@
+#!/bin/bash -ux
+
+model=`juju show-model --format=json| jq -r '.| keys[]'`
+machines=`juju status $model --format=json | jq -r '.machines | to_entries[].value."instance-id"'`
+num=`echo "$machines" | wc -l`
+
+for i in $(seq 1 $((num*2))); do
+	if [ "`openstack volume list --name kafka-vol-$i -c Status -f value`" != "available" ]; then
+		openstack volume create kafka-vol-$i --size 1
+	fi
+done
+
+i=1
+for machine in $machines; do
+	openstack server add volume $machine kafka-vol-$((i++))
+	openstack server add volume $machine kafka-vol-$((i++))
+done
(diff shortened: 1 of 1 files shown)


kafka/tools/create_volumes.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -ux
+
+model=`juju show-model --format=json| jq -r '.| keys[]'`
+machines=`juju status $model --format=json | jq -r '.machines | to_entries[].value."instance-id"'`
+num=`echo "$machines" | wc -l`
+
+for i in $(seq 1 $((num*2))); do
+	if [ "`openstack volume list --name kafka-vol-$i -c Status -f value`" != "available" ]; then
```
nicolasbock, line 8: If one of the volumes is already in use you would create another volume with the same name, which would lead to a failure in lines 15 and 16 because you are resolving the volume there by name and not ID.

kafka/tools/create_volumes.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -ux
+
+model=`juju show-model --format=json| jq -r '.| keys[]'`
+machines=`juju status $model --format=json | jq -r '.machines | to_entries[].value."instance-id"'`
```
nicolasbock, line 4: Do you intend to attach 2 volumes to every machine in the model? The title of the PR suggests that those are only necessary for Kafka units.

nicolasbock (changes requested):

kafka/tools/create_volumes.sh
drencrom: No, but I should run `juju status kafka` instead of using `$model`. It worked for me because my test model is called kafka also.

nicolasbock (changes requested):

kafka/tools/create_volumes.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash -ux
+
+machines=`juju status kafka --format=json | jq -r '.machines | to_entries[].value."instance-id"'`
+num=`echo "$machines" | wc -l`
+
+volumeIds=()
+for i in $(seq 1 $((num*2))); do
+	if [ "`openstack volume list --name kafka-vol-$i -c Status -f value`" != "available" ]; then
```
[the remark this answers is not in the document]
drencrom: What do you suggest to do in that case?

nicolasbock (approved):

PR #82, opened 2022-11-09
Fix error counting length of the commit subject

drencrom:
The command wc --chars <<<${subject} does not work right as it always counts one extra char. For example:

$ wc --chars <<<""
1

$ wc --chars <<<"12345678901234567890123456789012345678901234567890" 
51


nicolasbock (changes requested):
Thanks! Could you also fix this in the test for the commit message body?

nicolasbock (approved):
Thanks!

PR #84, opened 2022-11-10
Remove some unnecessary wc uses

pponnuvel:


nicolasbock (approved):
Nice! Looks much better :)

PR #86, opened 2022-12-02
Ensure default binding provided with maas

dosaboy:
When using maas it is mandatory to set a default
binding that exists (see LP #1994124).


sombrafam (approved):
LGTM.

PR #88, opened 2022-12-09
Add cinder-volume and cinder-purestorage overlays

lathiat:
Add support for deploying cinder-purestorage

Add support for deploying a separate cinder-volume service. Though not
strictly required in stsstack, this is to replicate customer
environments where the LXD cinder service cannot connect to iSCSI
volumes to clone images to them (as iSCSI is not namespaced), so a
separate cinder-volume service is deployed in KVM or on metal. Commonly
used with purestorage or other supported iSCSI SANs.


openstack/module_defaults
```diff
@@ -61,3 +61,5 @@ MOD_PARAMS[__DNS_DOMAIN__]="${model_subdomain}.stsstack.qa.1ss."
 MOD_PARAMS[__DESIGNATE_NAMESERVERS__]="ns1.${MOD_PARAMS[__DNS_DOMAIN__]}"
 
 MOD_PARAMS[__CONSOLE_TYPE__]="novnc"
+
```
dosaboy, line 64: nit: lose the whitespace

openstack/pipeline/02configure
```diff
@@ -635,6 +635,29 @@ do
             assert_min_release queens 'cinder-lvm'
             MOD_OVERLAYS+=( "cinder-lvm.yaml" )
             ;;
+        --cinder-volume)
+            assert_min_release ussuri 'cinder-volume'
+            MOD_OVERLAYS+=( "cinder-volume.yaml" )
+            MOD_PARAMS[__CINDER_VOLUME_INTERFACE__]='cinder-volume'
+            ;;
+        --cinder-purestorage)
+            assert_min_release ussuri 'cinder-volume'
```
dosaboy, line 644: i assume this should be:

assert_min_release ussuri 'cinder-purestorage'

openstack/pipeline/02configure
```diff
@@ -635,6 +635,29 @@ do
...
+            MOD_MSGS[cinder-purestorage.0]="For access to the serverstack purestorage SAN please contact the OpenStack team for the API Token / IP and specify them using --purestorage-api-token and --purestorage-san-ip"
+            MOD_MSGS[cinder-purestorage.1]="Volumes created on the PureStorage SAN are not easily identifiable - you cannot see which user created them and they all have the name volume-%s. Please ensure you delete all volumes before destroying this environment to prevent orphaned volumes accumulating."
+            MOD_MSGS[cinder-purestorage.2]="After creation create volume types like this:"
+            MOD_MSGS[cinder-purestorage.3]="openstack volume type create cinder-purestorage --property volume_backend_name=cinder-purestorage"
+            MOD_MSGS[cinder-purestorage.4]="openstack volume type create cinder-ceph --property volume_backend_name=cinder-ceph # If also deploying ceph"
+            MOD_MSGS[cinder-purestorage.5]="juju config cinder default-volume-type=cinder-ceph # or cinder-purestorage"
+            ;;
+        --purestorage-api-token)  #__OPT__type:<str>
+            MOD_PARAMS[__PURESTORAGE_API_TOKEN__]=$2
+            shift
```
dosaboy, line 655: If i omit these option i get an error like:

ERROR: there are unrendered variables in your bundle:
./b/o/cinder-purestorage.yaml:      pure-api-token: __PURESTORAGE_API_TOKEN__
./b/o/cinder-purestorage.yaml:      san-ip: __PURESTORAGE_SAN_IP__

Another way to solve this is to use get_param e.g.

```
        msg="REQUIRED: purestorage api token"
        get_param $1 __PURESTORAGE_API_TOKEN__ "$msg"
        msg="REQUIRED: purestorage san ip:"
        get_param $1 __PURESTORAGE_SAN_IP__ "$msg"
```

lathiat: This didn't let you still use --purestorage-san-ip as well (it seems?) and you'd have to type them in manually every time. So I just added MOD_DEFAULTS instead so it renders them blank but still has the MOD_MSGS warning. Let me know if that works for you.

lathiat:
Address comments

overlays/cinder-volume.yaml
```diff
@@ -0,0 +1,33 @@
...
+      block-device:
+      glance-api-version: 2
+      openstack-origin: *openstack_origin
+      ssl_ca: *ssl_ca
+      ssl_cert: *ssl_cert
+      ssl_key: *ssl_key
+relations:
+  - [ cinder-volume:shared-db, __MYSQL_INTERFACE__ ]
+  - [ cinder-volume, rabbitmq-server ]
+# Add when https://review.opendev.org/c/openstack/charm-cinder/+/845558 is merged
```
dosaboy, line 32: the mentioned patch is now merged so i guess you need to uncomment this?

dosaboy:
lgtm but there is one small thing that needs fixing it seems (have left inline comment)

overlays/cinder-volume.yaml
lathiat: I might hold off on merging that until the patch that makes that works is backported all the way to ussuri at least. Or otherwise we have to try add some kind of conditional currently it only works in Yoga+ and only on edge not stable right now I think.

Thoughts?

dosaboy (approved):
this resolves current issues, we can update as and when backports land

PR #92, opened 2023-02-21
Cinder Improvements

lathiat:
Move LVM configuration from cinder to cinder-lvm charm
Use storage spaces to allocate storage for cinder-lvm
Use lioadm for cinder-lvm (bug workaround)
Use cinder-volume:identity-credentials by default

### DIFF
--- a/overlays/cinder-volume.yaml
+++ b/overlays/cinder-volume.yaml
@@ -20,7 +20,7 @@ applications:
       debug: *debug
       verbose: *verbose
       enabled-services: volume
-      block-device:
+      block-device: None
       glance-api-version: 2
       openstack-origin: *openstack_origin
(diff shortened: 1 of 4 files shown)


nicolasbock (approved):
Thanks @lathiat ! LGTM

overlays/cinder-volume.yaml
```diff
@@ -20,7 +20,7 @@ applications:
       debug: *debug
       verbose: *verbose
       enabled-services: volume
-      block-device:
+      block-device: None
```
dosaboy, line 23: this will give the string "None", is that expected? In yaml None is null.

dosaboy: actually that might not be the case for the Juju yaml parser, need to check

PR #95, opened 2023-03-23
Fix target helper for victoria series

sombrafam:
The target_helper lioadm was only introduced as default in wallaby. So, older versions don't have it. See
https://bugs.launchpad.net/charm-cinder-lvm/+bug/1949074 for more info


sombrafam:
@rodrigogansobarbieri You complained about the LVM errors in Cinder.
@lathiat This kind of a re-write of your patch but only applying it for wallaby and above.

dosaboy (approved):
lgtm

PR #98, opened 2023-04-27
OVN debug only supported for >= yoga

dosaboy:


openstack/pipeline/02configure
```diff
@@ -477,6 +477,9 @@ do
             # since ports typically set post-deployment ensure we don't clobber on re-run
             opt=`update_from_charm_config $1 ovn-chassis bridge-interface-mappings`
             MOD_OVERLAYS+=( "neutron-ovn.yaml" )
+            if has_min_release yoga; then
```
brianphaley, line 480: I haven't tested this but from looking at the code it might not work.

has_min_release() calls assert_min_release() which will 'exit 1' if the condition is not met. So there might be some refactor necessary.

dosaboy: I've just realised that in fact the ovn-central charm never had a debug config option so the actual fix is just to remove that completely. I will update the patch.

dosaboy: to answer your question about has_min_release(), the exit value of assert_min_release is used as the return value of has_min_release i.e. the exit happens in the context of the subprocess $(...) so does not cause the caller to exit.

brianphaley (approved):
And thanks for the info on the 'exit 1' code, learned something new about shell :)

brianphaley (approved):

PR #99, opened 2023-05-10
ch_channel_map: Fixes & updates to charm versions

lathiat:
The main change here is that bionic-ussuri used to be a symlink to focal
and focal-yoga was a symlink to jammy. This meant incorrectly using
the wrong versions of the charm for hacluster, rabbitmq, etc.

While they worked, this matters when testing a series-upgrade as in
reality you need to first upgrade these charms before the
series-upgrade. It also meant we weren't replicating the charm version
likely to be in the customer environment when not upgrading. This
resulted in not noticing this critical bug before asking a customer to
series upgrade: https://bugs.launchpad.net/juju/+bug/2008509

Additionally updated all of the charm versions to be aligned with the
current "Charm Delivery" suggestions with two exceptions
- Vault: Field still deploys Jammy with vault 1.7, replicate that to
  match customer deployments better.
- OVN: OVN 22.03 is now recommended for all OpenStack releases
  Ussuri-Yoga, when previously they had different matching OVN versions.
  Most customers on versions older than Yoga still have the older OVN
  version, so I kept those to make it more likely we replicate customer
  problems since OVN is the source of many such problems. We probably need
  to add an option to optionally select OVN 22.03 for these versions.

Barbican-vault used to only have a Yoga+Zed channel at some point
but this has been fixed, so the special exception for this was removed.

Fix Wallaby to use Ceph pacific


dosaboy (approved):
looks good, thanks for doing this

PR #100, opened 2023-05-10
mysql-innodb-cluster: no SSL by default

lathiat:
Don't use vault certificates by default with MySQL InnoDB Cluster.

The certificates aren't setup until you unlock the vault, which you
can't do until halfway through the charm deployment after MySQL has
already started since vault itself depends on MySQL.

When you unlock it, the certificates are created and the MySQL charm
then reconfigures and restarts MySQL which causes any OpenStack service
that is doing database migration/setup to fail and error out
permanently.

In my experience, this happens most of the time in practice.

We're not normally using this for real deployments, so add the
--mysql-innodb-cluster-ssl option to opt-in instead.

### DIFF
--- a/openstack/pipeline/02configure
+++ b/openstack/pipeline/02configure
@@ -371,15 +371,15 @@ do
                 MOD_OVERLAYS+=( "vault-openstack-certificates-masakari.yaml" )
             fi
             ;;
+        --mysql-innodb-cluster-ssl)
+            MOD_OVERLAYS+=( "vault-certificates-mysql-innodb-cluster.yaml" )
+            ;;
         --mysql-innodb-cluster*)
             conflicts_with $1 --mysql-ha*
             assert_min_release train mysql-innodb-cluster
             # NOTE: 3 is actually the absolute minimum units required
             get_units $1 __NUM_MYSQL_UNITS__ 3
             MOD_OVERLAYS+=( "mysql-innodb-cluster.yaml")
-            if has_opt --vault; then
-                MOD_OVERLAYS+=( "vault-certificates-mysql-innodb-cluster.yaml" )
-            fi
             # this will be auto-generated for each app (see common/render.d/all)
             MOD_OVERLAYS+=( "mysql-innodb-cluster-router.yaml" )
             MOD_PARAMS[__MYSQL_INTERFACE__]='__APPLICATION_MYSQL_INNODB_ROUTER__'


lathiat:
@dosaboy specifically need some input from you on this one.. trying to figure out why we had '--mysql-innodb-cluster*' with an asterisk.. I can't see a reason. It interfered with adding the option mysql-innodb-cluster-ssl.. so I removed it.

But I am not sure if I am totally missing some reason this is there? Some options seem to use it at random. Like barbican-ha, ceph-fs-ha, etc. Am I missing something?

nicolasbock:
That is weird @lathiat. I can't think of a reason either.  @dosaboy?

dosaboy:
sorry i never came back to this. so the overlays with an asterisk are ones for whom a number of units can be specified:

```
$ ./generate-bundle.sh --list-overlays| egrep "\*$"
--barbican-ha*
--ceph-fs-ha*
--ceph-rgw-ha*
--cinder-ha*
--designate-ha*
--dvr-l3ha*
--dvr-snat*
--dvr-snat-l3ha*
--glance-ha*
--ha*
--heat-ha*
--keystone-ha*
--l3ha*
--manila-ganesha-ha*
--mysql-ha*
--mysql-innodb-cluster*
--neutron-api-ha*
--nova-cloud-controller-ha*
--nova-console*
--octavia-ha*
--openstack-dashboard-ha*
--placement-ha*
--rabbitmq-server-ha*
--swift-ha*
--telemetry-ha*
--vault-ha*
```

so e.g. you can do --mysql-innodb-cluster:10 if you want 10 units of mysql-innodb-cluster

zhhuabj:
Today I had planed to set up a test env quickly by './generate-bundle.sh -s focal --name focal --num-compute 3 --vault --use-stable-charms --run' in stsstack, but neutron-api/0 failed to run 'neutron-db-manage --config-file /etc/neutron/neutron.conf --config-file /etc/neutron/plugins/ml2/ml2_conf.ini upgrade head' with the following error:

oslo_db.exception.DBNonExistentTable: (pymysql.err.InternalError) (1051, "Unknown table 'neutron.cisco_n1kv_port_bindings'")

Finally it was fixed by this unmerged PR. thanks Trent.

lathiat:
> sorry i never came back to this. so the overlays with an asterisk are ones for whom a number of units can be specified:
> 
> ```
> $ ./generate-bundle.sh --list-overlays| egrep "\*$"
> --barbican-ha*
> --ceph-fs-ha*
> --ceph-rgw-ha*
> --cinder-ha*
> --designate-ha*
> --dvr-l3ha*
> --dvr-snat*
> --dvr-snat-l3ha*
> --glance-ha*
> --ha*
> --heat-ha*
> --keystone-ha*
> --l3ha*
> --manila-ganesha-ha*
> --mysql-ha*
> --mysql-innodb-cluster*
> --neutron-api-ha*
> --nova-cloud-controller-ha*
> --nova-console*
> --octavia-ha*
> --openstack-dashboard-ha*
> --placement-ha*
> --rabbitmq-server-ha*
> --swift-ha*
> --telemetry-ha*
> --vault-ha*
> ```
> 
> so e.g. you can do --mysql-innodb-cluster:10 if you want 10 units of mysql-innodb-cluster

OK I see. Thanks for the clarification. I don't see anyone using this in practice, so I wonder if we can safely drop it here in favour of the SSL option?

I think this PR is much more needed than the rare case of someone wanting to deploy with a custom number of MySQL units.

Or otherwise can we somehow have both work?


I hit this issue a lot and as you saw, Joshua also hit it today.

nicolasbock (changes requested):
You don't need to get rid of the `*` syntax if you switch the order of the 2 `--mysql*` cases. In fact you could reduce this PR to
```diff
diff --git a/openstack/pipeline/02configure b/openstack/pipeline/02configure
index 05d4a44..4fb0360 100644
--- a/openstack/pipeline/02configure
+++ b/openstack/pipeline/02configure
@@ -325,15 +325,15 @@ do
                 MOD_OVERLAYS+=( "vault-openstack-certificates-masakari.yaml" )
             fi
             ;;
+        --mysql-innodb-cluster-ssl)
+            MOD_OVERLAYS+=( "vault-certificates-mysql-innodb-cluster.yaml" )
+            ;&
         --mysql-innodb-cluster*)
             conflicts_with $1 --mysql-ha*
             assert_min_release train mysql-innodb-cluster
             # NOTE: 3 is actually the absolute minimum units required
             get_units $1 __NUM_MYSQL_UNITS__ 3
             MOD_OVERLAYS+=( "mysql-innodb-cluster.yaml")
-            if has_opt --vault; then
-                MOD_OVERLAYS+=( "vault-certificates-mysql-innodb-cluster.yaml" )
-            fi
             # this will be auto-generated for each app (see common/render.d/all)
             MOD_OVERLAYS+=( "mysql-innodb-cluster-router.yaml" )
             MOD_PARAMS[__MYSQL_INTERFACE__]='__APPLICATION_MYSQL_INNODB_ROUTER__'
```

lathiat:
Updated, thanks for the fix @nicolasbock :)

nicolasbock (approved):
Thanks @lathiat

PR #101, opened 2023-05-10
neutron-ext-net: Customise physical network

lathiat:
Give neutron-ext-net the ability to choose which physical network you
want to use.

Useful when replicating environments with multiple physical
networks including SR-IOV bonding environments.


dosaboy (approved):
lgtm

PR #102, opened 2023-05-10
Fix ha overlays

tpsilva:
vip_iface is not being set, so it defaults to eth0, which fails on all ha openstack deployments. Setting it to default to ens3 fixes that.


dosaboy:
this has never failed for me fwiw, i wonder why

nicolasbock:
What OS series did you use @tpsilva?

nicolasbock:
I cannot reproduce with
```console
./generate-bundle.sh --release ussuri --ovn --ha
```
Am I missing something?

nicolasbock:
You have more than `10.5.0.0/16` alpha space?

tpsilva:
Yes, I had another subnet that I added to reproduce a Juju issue. Removing that subnet fixed the HA issue...

PR #103, opened 2023-05-15
Fix Kubernetes deployments

tpsilva:
Kubernetes deployments were failing because it was using the old charms from containers-*. This patch removes that and changes references from kubernetes-master to kubernetes-control-plane, which has been updated on 1.24[1].

[1] https://ubuntu.com/kubernetes/docs/release-notes#124


nicolasbock (changes requested):
Please also rename the `k8s-master-*` template files to match the transition from `master` to `control-plane`?

nicolasbock (changes requested):
You also need to change https://github.com/canonical/stsstack-bundles/blob/93c1034c483348c43292f2dc361ab6eb284bec39/kubernetes/pipeline/02configure#L171

tpsilva:
Will do that and test to make sure all is correct... new patch coming in a bit

tpsilva:
Fixed all references to kubernetes-master. Tested with 3 control planes.

dosaboy:
lgtm, please be sure to test a deployment with this before we land it

tpsilva:
Tested with:

`$ ./generate-bundle.sh --num-control-planes 3 --name k8s`

Deployment works as expected

nicolasbock (approved):
Thanks!

PR #105, opened 2023-05-18
Relax --network syntax for instance creation

nicolasbock:

### DIFF
--- a/openstack/tools/instance_launch.sh
+++ b/openstack/tools/instance_launch.sh
@@ -7,7 +7,8 @@ set -e -u
 # Presumes glance images exist and have been imported using the accompanying
 # configure script.
 
-declare -a net_names=(private)
+declare -a DEFAULT_NETS=(private)
+declare -a net_names=()
 instance_qty=0
 image_name=
 server_name=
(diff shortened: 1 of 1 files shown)


openstack/tools/instance_launch.sh
```diff
@@ -49,8 +50,14 @@ EOF
...
             IFS=','
-            read -r -a net_names <<< "$1"
+            read -r -a new_net_names <<< "$1"
             IFS=${old_IFS}
+            if ((net_names_set == 0)); then
+                net_names=( ${new_net_names[@]} )
+                net_names_set=1
+            else
+                net_names=( ${net_names[@]} ${new_net_names[@]} )
+            fi
```
brianphaley, line 60: Just curious if you could have used += here, for example:

net_names+=(" ${new_net_names[@]}")

Untested of course.

nicolasbock: Thanks, that should work :smile:

openstack/tools/instance_launch.sh
```diff
@@ -49,8 +50,14 @@ EOF
...
             IFS=','
-            read -r -a net_names <<< "$1"
+            read -r -a new_net_names <<< "$1"
             IFS=${old_IFS}
+            if ((net_names_set == 0)); then
+                net_names=( ${new_net_names[@]} )
+                net_names_set=1
+            else
+                net_names+=( ${new_net_names[@]} )
+            fi
```
brianphaley, line 60: I think if you define net_names = "" above you don't even need this if/else logic, just keep adding to it

PR #108, opened 2023-05-26
Handle previously invalid vault CA

lathiat:
If you run novarc/install_local_ca.sh when vault is not yet initialised,
the previously invalid value of None was stored and causes future calls
to these scripts to error out.

While the script checked for this, after discovering the None it would
create a second temporary file in /tmp matching the "find" query and
when you call the script again in future in finds both files which
causes the script to error out.

There is not much point in using 'mktemp' to get a random path and then
trying to find it again using a unique part of the filename, so switch
to using a static path in /tmp which avoids having more than 1 match to
the path.

Secondly rather than check for the "None" error, more explicitly check
we get an actual SSL certificate back and if not, try to fetch the CA
again. Try to also remove such invalid output so it's not re-used in
future but still always check for it in case that doesn't happen for
some reason.

### DIFF
--- a/openstack/tools/install_local_ca.sh
+++ b/openstack/tools/install_local_ca.sh
@@ -1,13 +1,27 @@
 #!/bin/bash -eu
-model_ca_cert_path=${1:-}
+
+validate_or_remove_ca() {
+    if [ -f $1 ]; then
+        if grep -Fxq -- "-----BEGIN CERTIFICATE-----" $1; then
+            return 0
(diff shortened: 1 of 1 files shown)


lathiat:
The commit message linter is failing due to a bug in the linter, submitted a fix here:
https://github.com/canonical/stsstack-bundles/pull/109

nicolasbock (approved):

PR #109, opened 2023-05-26
lint-git-messages.sh: Fix line width detection

lathiat:
The body line width detector was using the '>' comparison which is meant
for strings, not integers. In practice it was failing on a 9 character
wide line.

Switch to the correct "-gt" integer check.


lathiat:
The commit that was failing is here: https://github.com/canonical/stsstack-bundles/actions/runs/5088497010/jobs/9144973861?pr=108

dosaboy (approved):

dosaboy (approved):

nicolasbock:
Good catch! These comparison subtleties are tricky.

PR #110, opened 2023-06-07
Fix hacluster channel

nicolasbock:


nicolasbock:
```console
$ juju deploy --channel 2.0.3/edge hacluster
ERROR resolving retry error: No revision was found in the Store.
```

does not work, but

```console
$ juju deploy --series focal --channel 2.0.3/edge hacluster
Located charm "hacluster" in charm-hub, revision 113
Deploying "hacluster" from charm-hub charm "hacluster", revision 113 in channel 2.0.3/stable on focal
```

does. There is no `default-series` in the model but I don't think that that's the issue. Not specifying the series works for other charms.

nicolasbock:
Thanks for the suggestion @rodrigogansobarbieri . I will rework this patch.

nicolasbock:
@rodrigogansobarbieri Please have another look.

rodrigogansobarbieri:
tested it locally, before applying the patch it didn't work, after applying it worked! LGTM. Very simple and efficient solution Nick!

nicolasbock:
Thanks @rodrigogansobarbieri !

Issue #112, opened 2023-07-04
Add Lunar/Antelope support


dnegreira:
We need to update our scripts in order to be able to deploy Lunar/Antelope

PR #113, opened 2023-07-05
Add lunar/antelope support

dnegreira:
Due to some checks being done by alphabetical order, it would break how we would assign and ensure that minimal versions of certain charms are installed in certain openstack versions.
How I approach the fix is by putting all the relevant openstack versions in an ordered array so that we can compare numerically the openstack versions between eachother.

Closes: #112

### DIFF
--- a/common/ch_channel_map/any_series
+++ b/common/ch_channel_map/any_series
@@ -1,5 +1,9 @@
+# newer openstack release names map to a release number
+declare -A RELNAME_MAP=(
+    [antelope]=2023.1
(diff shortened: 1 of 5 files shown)


dosaboy:
there are a few things that dont quite work as they should. i have gone ahead and made some changes myself and here is the diff that you can apply to yours - https://pastebin.ubuntu.com/p/WJRz9cKwGk/

dosaboy (approved):
lgtm!

Issue #121, opened 2023-08-22
default to ml2-ovn for ussuri and up


dosaboy:
we currently default to ml2-ovs for all versions of openstack but ml2-ovn is the most commonly used these days so i think its time to switch the default

PR #122, opened 2023-08-22
Default to OVN for release >= Ussuri

nicolasbock:
Closes: https://github.com/canonical/stsstack-bundles/issues/121

### DIFF
--- a/openstack/pipeline/02configure
+++ b/openstack/pipeline/02configure
@@ -28,6 +28,11 @@ else
     MOD_PARAMS[__NUM_NEUTRON_GATEWAY_UNITS__]=0
 fi
 
+# Default to OVN for > Train
+if has_min_release ussuri && ! (has_opt --ovs || has_opt --ml2-ovs); then
+    set -- $@ --ml2-ovn && cache $@
+fi
+
 if ! has_opt --ovn*; then
     MOD_OVERLAYS+=( "neutron-openvswitch.yaml" )
 fi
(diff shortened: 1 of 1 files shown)


openstack/pipeline/02configure
```diff
@@ -483,6 +488,8 @@ do
                 set -- $@ --vault && cache $@
             fi
             ;;
+        --ovs)
```
dosaboy, line 491: i think this is ok but maybe we should call it --ml2-ovs

nicolasbock: Should we also rename `--ovn` to `--ml2-ovn` then to make this option more
"symmetric"? This would cause a behavior change though, so maybe not a good
idea.

nicolasbock: Ok, will do that.

openstack/pipeline/02configure
```diff
@@ -28,6 +28,11 @@ else
     MOD_PARAMS[__NUM_NEUTRON_GATEWAY_UNITS__]=0
 fi
 
+# Default to OVN for > Train
+if has_min_release ussuri && ! has_opt --ovs; then
+    set -- $@ --ovn && cache $@
```
brianphaley, line 33: nit: --ml2-ovn ?

Also, the commit message doesn't mention anything about adding the option, or the new --ml2-ovs option either.

openstack/pipeline/02configure
```diff
@@ -28,6 +28,11 @@ else
     MOD_PARAMS[__NUM_NEUTRON_GATEWAY_UNITS__]=0
 fi
 
+# Default to OVN for > Train
+if has_min_release ussuri && ! has_opt --ovs; then
```
brianphaley, line 32: --ml2-ovs, or an alias should be added below

brianphaley (changes requested):

openstack/pipeline/02configure
nicolasbock: Good catch.

openstack/pipeline/02configure
nicolasbock: Changed. I will update the commit message.

dosaboy:
lgtm thanks nick!

dosaboy (approved):

PR #127, opened 2023-08-24
Remove nova-network support

brianphaley:
"It's dead Jim"

This does not remove any of the Rally templates associated with the NovaNetworks class in rally-openstack.

Trivialfix


brianphaley:
Unless we really do have a customer running nova-network...

nicolasbock (approved):
I am not aware of any customer using this. @dosaboy ?

Issue #131, opened 2023-09-11
deploying bionic ussuri uses latest/stable for some charms


dosaboy:
While this used to be valid it is no longer since that channel has been updated to be more recent version that does not support bionic.

lathiat:
rabbitmq is one of those charms with some jacked up release info in `juju info rabbitmq-server --format json | jq`

3.6/edge r123 supports 16.04, 18.04, 20.04, 20.10, 21.04, 21.10
3.6/edge r133 supports 16.04, 18.04
3.6/edge r135 supports 16.04, 20.04

and there is no 3.6/stable

so 3.6/edge might work
otherwise 3.8/stable probably also will



ovn also needs to be set to 20.03

nicolasbock:
Closing since fixed in #132

PR #132, opened 2023-09-11
Update bionic versions

dosaboy:
OVN and Rabbitmq charm channels needed adjusting
to use versions that support Bionic.

Closes-Bug: #131


brianphaley (approved):
Looks to match Trent's comments.

nicolasbock (approved):

PR #146, opened 2023-10-26
Remove hard-coded unit reference

nicolasbock:
When keystone unit 0 is missing or another unit is the current leader,
the hard-coded reference to `keystone/0` will yield a null or incorrect
result.


dosaboy (approved):
lgtm

Issue #149, opened 2023-10-30
cannot deploy generated bundle in juju 3.x


rafalop:
Tested juju 3.2 only. It seems that you can no longer add relations into a bundle if the relation involves an application deployed by an overlay. This is the result:

```
(os) ubuntu@bastion:~/stsstack-bundles/openstack$ juju deploy /home/ubuntu/stsstack-bundles/openstack/b/zedmodel/openstack.yaml --overlay /home/ubuntu/stsstack-bundles/openstack/b/zedmodel/o/openstack/glance.yaml --overlay /home/ubuntu/stsstack-bundles/openstack/b/zedmodel/o/openstack/keystone.yaml --overlay /home/ubuntu/stsstack-bundles/openstack/b/zedmodel/o/openstack/neutron-ovn.yaml --overlay /home/ubuntu/stsstack-bundles/openstack/b/zedmodel/o/openstack/neutron-ml2dns.yaml --overlay /home/ubuntu/stsstack-bundles/openstack/b/zedmodel/o/mysql-innodb-cluster.yaml --overlay /home/ubuntu/stsstack-bundles/openstack/b/zedmodel/o/mysql-innodb-cluster-router.yaml --overlay /home/ubuntu/stsstack-bundles/openstack/b/zedmodel/o/openstack/placement.yaml --overlay /home/ubuntu/stsstack-bundles/openstack/b/zedmodel/o/vault.yaml --overlay /home/ubuntu/stsstack-bundles/openstack/b/zedmodel/o/openstack/vault-openstack-secrets.yaml --overlay /home/ubuntu/stsstack-bundles/openstack/b/zedmodel/o/openstack/vault-openstack-certificates.yaml --overlay /home/ubuntu/stsstack-bundles/openstack/b/zedmodel/o/openstack/vault-openstack-certificates-placement.yaml
ERROR cannot deploy bundle: the provided bundle has the following errors:
relation ["nova-cloud-controller:shared-db" "nova-cloud-controller-mysql-router:shared-db"] refers to application "nova-cloud-controller-mysql-router" not defined in this bundle
relation ["nova-cloud-controller" "glance"] refers to application "glance" not defined in this bundle
relation ["nova-cloud-controller" "keystone"] refers to application "keystone" not defined in this bundle
relation ["nova-compute" "glance"] refers to application "glance" not defined in this bundle
relation ["glance" "cinder:image-service"] refers to application "glance" not defined in this bundle
[...]

PR #156, opened 2023-11-21
Separate relations into separate yaml document

rafalop:
To adjust to new juju 3.2 behaviour that does not
allow relations referencing overlay applications
in the base bundle.

Ref: https://bugs.launchpad.net/juju/+bug/2041785

Resolves: #149

### DIFF
--- a/tools/juju-bundle-applications.py
+++ b/tools/juju-bundle-applications.py
@@ -7,8 +7,8 @@ application_list = set()
 
 for filename in sys.argv[1:]:
     with open(filename, 'r') as f:
-        data = yaml.load(f, Loader=yaml.SafeLoader)
-        if 'applications' in data:
-            application_list.update(data['applications'].keys())
-
+        data = yaml.load_all(f, Loader=yaml.SafeLoader)
+        for d in data:
+            if 'applications' in d:
+                application_list.update(d['applications'].keys())
 print('\n'.join(application_list))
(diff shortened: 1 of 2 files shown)


rafalop:
FYI this change is compatible with older juju versions (ie 2.x).

tools/juju-bundle-applications.py
```diff
@@ -7,7 +7,7 @@
 
 for filename in sys.argv[1:]:
     with open(filename, 'r') as f:
-        data = yaml.load(f, Loader=yaml.SafeLoader)
+        data = yaml.load_all(f, Loader=yaml.SafeLoader)
         if 'applications' in data:
             application_list.update(data['applications'].keys())
```
nicolasbock, line 13: The `yaml.load_all()` function returns a list of documents:
```suggestion
        for d in data:
            if 'applications' in d:
                application_list.update(d['applications'].keys())
```

nicolasbock (changes requested):
Neat that the fix is so simple :smile:

nicolasbock (approved):

lathiat:
I only just caught up on this one and just wanted to note that using --- to start a new yaml document within the same file is a fantastic hack. 

I had solved this in my local branch by deploying an overlay/blank.yaml as the base bundle, and then passing in the real base bundle (e.g. openstack.yaml) as the first overlay.

PR #160, opened 2023-11-30
Add initial set of tarreform plans

arif-ali:
An initial start for testing purposes


nicolasbock:
Could you also include a short README or something similar describing how to use these plans?

PR #165, opened 2023-12-21
Add a configure script to JAAS

drencrom:
The script finishes the configuration after the model has been deployed


nicolasbock (approved):
Thanks!

PR #170, opened 2024-01-29
Fix testkey hint in instance_launch.sh

brianphaley:
It incorrectly says to use ~/testkey.pem when it should be ~/testkey.priv.

TrivialFix


nicolasbock (approved):
Indeed :smile:

PR #172, opened 2024-02-21
Change systemd-resolve for resolvectl

drencrom:
The `systemd-resolve` tool has been renamed to `resolvectl` in [systemd 239](https://github.com/systemd/systemd/blob/main/NEWS#L8539). 
From Jammy onwards the symlink `systemd-resolve` has been deleted and this code does not work.
Focal has systemd 245 so this change in fact removes support for Bionic and adds support for Jammy and following releases. If support for Bionic is still desired the patch should be expanded.

### DIFF
--- a/common/helpers
+++ b/common/helpers
@@ -825,7 +825,13 @@ print_msgs ()
 
 get_local_upstream_dns ()
 {
-    local server="`systemd-resolve --status 2>/dev/null| sed -r 's/\s+DNS Servers:\s+([[:digit:]]+)/\1/g;t;d'| head -n 1`"
+    # Use old resolve command if present
+    if [ -f /usr/bin/systemd-resolve ]; then
+        local resolver="systemd-resolve --status"
+    else
+        local resolver="resolvectl"
+    fi
+    local server="`$resolver 2>/dev/null| sed -r 's/\s+DNS Servers:\s+([[:digit:]]+)/\1/g;t;d'| head -n 1`"
     (($?==0)) && echo $server && return 0
     echo "10.198.200.1"  # stsstack upstream dns
 }
--- a/common/helpers
+++ b/common/helpers
@@ -825,14 +825,13 @@ print_msgs ()
 
 get_local_upstream_dns ()
 {
-    # Use old resolve command if present
-    if [ -f /usr/bin/systemd-resolve ]; then
-        local resolver="systemd-resolve --status"
+    # Use old systemd-resolve command if present
+    if command -v systemd-resolve > /dev/null; then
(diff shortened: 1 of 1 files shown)


nicolasbock:
Thanks for the fix!

We are supporting Bionic / Queens until April 2028. We need to be able to deploy on Bionic until then. Could you expand this change so that Bionic still works? Thanks!

drencrom:
Added check for `systemd-resolve` so it works with versions of systemd older than 239.

nicolasbock (approved):
Thanks!

common/helpers
```diff
@@ -825,7 +825,13 @@ print_msgs ()
...
-    local server="`systemd-resolve --status 2>/dev/null| sed -r 's/\s+DNS Servers:\s+([[:digit:]]+)/\1/g;t;d'| head -n 1`"
+    # Use old resolve command if present
+    if [ -f /usr/bin/systemd-resolve ]; then
+        local resolver="systemd-resolve --status"
+    else
+        local resolver="resolvectl"
+    fi
+    local server="`$resolver 2>/dev/null| sed -r 's/\s+DNS Servers:\s+([[:digit:]]+)/\1/g;t;d'| head -n 1`"
     (($?==0)) && echo $server && return 0
     echo "10.198.200.1"  # stsstack upstream dns
```
nicolasbock, line 836:
```suggestion
    # Use old systemd-resolve command if present
    if command -v systemd-resolve > /dev/null; then
        local server=$(systemd-resolve --status 2> /dev/null | sed --regexp-extended 's/\s*DNS Servers:\s+([[:digit:]]+)/\1/g;t;d'| head -n 1)
    else
        local server=$(resolvectl 2> /dev/null | sed --regexp-extended 's/\s*Current DNS Server:\s+([[:digit:]]+)/\1/g;t;d')
    fi
    (($? == 0)) && echo $server && return 0
    echo "10.198.200.1"  # stsstack upstream dns
```

nicolasbock: After more careful testing I realized that the output format is slightly different between the two commands.

nicolasbock (changes requested):

nicolasbock:
> Thanks for the fix!
> 
> We are supporting Bionic / Queens until April 2028. We need to be able to deploy on Bionic until then. Could you expand this change so that Bionic still works? Thanks!

Just forget what I said there. It has nothing to do with your change :/

I would still like to keep support for Bionic, thanks for all of the updates!

dosaboy (approved):
lgtm

nicolasbock (approved):
Thanks!

PR #175, opened 2024-03-14
Fix few issues to make microceph script work

pponnuvel:


openstack/tools/create-microceph-vm.sh
```diff
@@ -107,8 +112,9 @@ done
 echo "done"
 
 # Add storage
-for i in $(seq 3); do
-    if ! lxc storage volume show default osd-${i}; then
+for i in {1..3}; do
+    if ! lxc storage volume show default osd-${i} 2>/dev/null; then
```
pponnuvel, line 116: Random "error" messages from scripts worry me, so sending it oblivion ;-)

openstack/tools/create-microceph-vm.sh
```diff
@@ -166,21 +169,21 @@ TOKEN_3=$(as_root ${VM_NAME}-1 microceph cluster add ${VM_NAME}-3)
 as_root ${VM_NAME}-2 microceph cluster join ${TOKEN_2}
 as_root ${VM_NAME}-3 microceph cluster join ${TOKEN_3}
 
-as_user ${VM_NAME}-1 microceph.ceph status
+as_root ${VM_NAME}-1 ceph status
```
pponnuvel, line 172: Keyrings are all owned by root. So ubuntu user can't run these commands.

openstack/tools/create-microceph-vm.sh
```diff
@@ -166,21 +169,21 @@ TOKEN_3=$(as_root ${VM_NAME}-1 microceph cluster add ${VM_NAME}-3)
...
-for i in $(seq 3); do
+for i in {1..3}; do
     as_root ${VM_NAME}-${i} microceph disk add /dev/sdb --wipe
 done
 
 sleep 2
 
-as_user ${VM_NAME}-1 microceph.ceph status
-as_user ${VM_NAME}-1 microceph.ceph osd status
+as_root ${VM_NAME}-1 ceph status
```
pponnuvel, line 181: Microceph snap has `ceph` alias for `microceph.ceph`.

nicolasbock (approved):
Thanks @pponnuvel ! Everything looks good to me. Just one question: Why do you prefer `{1..3}` over `$(seq 3)`?

pponnuvel:
> Thanks @pponnuvel ! Everything looks good to me. Just one question: Why do you prefer `{1..3}` over `$(seq 3)`?

I prefer to avoid subshells (fork/exec) whenever possible.

nicolasbock:
Thanks, that's a good point. That's why I like to use `[[` instead of `[`.

PR #184, opened 2024-05-06
Fix juju version check

drencrom:
Previously it didn't work for juju versions greater that 3.2

### DIFF
--- a/jaas/pipeline/02configure
+++ b/jaas/pipeline/02configure
@@ -5,7 +5,8 @@
 # All overlay/bundle variables (MOD_PARAMS) defaults must go into
 # the <module>/module_defaults file.
 
-if ! `juju --version| egrep -q "^3.2"`; then
+IFS='-'; juju_ver=($(juju --version)); unset IFS;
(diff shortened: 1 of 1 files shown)


dosaboy (approved):
lgtm

jaas/pipeline/02configure
```diff
@@ -5,7 +5,8 @@
 # All overlay/bundle variables (MOD_PARAMS) defaults must go into
 # the <module>/module_defaults file.
 
-if ! `juju --version| egrep -q "^3.2"`; then
+IFS='-'; juju_ver=($(juju --version)); unset IFS;
```
pponnuvel, line 8: Sorry I am a bit late here.

I do see one problem here. The typical way, when changing `IFS`, is to restore the old `IFS`.
Otherwise it'll break *word splitting* in the rest of the script.

But this could simplified without having to restore IFS if written like:
```
IFS='.' read -r major minor ignore < <(juju --version)
if [[ $major -ge 3 && $minor -ge 2 ]]; then
```

which is easier to understand as well (imo).

PR #187, opened 2024-05-14
Tweak juju version check

pponnuvel:


dosaboy (approved):
i'll take your word for it that this works ;)

PR #194, opened 2024-05-24
Use 'openstack subnet list' to retrieve VIP subnets

brianphaley:
Using 'openstack subnet show' on a subnet that does not exist will return an error, instead use 'subnet list' with --name which will just return an empty string if it is not found.


nicolasbock (approved):
Thanks, definitely cleaner :smile:

PR #199, opened 2024-06-12
Fix SWIFT_IP for prodstack6

nicolasbock:
On PS6 we have to use TLS for talking to radosgw (Swift). This change
updates the use of `SWIFT_IP` everywhere so that the protocol, http or
https, can be specified.


nicolasbock:
Not needed.

PR #204, opened 2024-07-11
Adds proxy and docker configuration for PS6

sombrafam:


kubernetes/configure
```diff
@@ -1,4 +1,11 @@
-#!/bin/bash -x
+#!/bin/bash
+
+profile=${1:-stsstack}
+
+if [[ $profile == "stsstack" ]]; then
```
dnegreira, line 5: We can drop these checks since stsstack is going away by the end of the week, lets make it compatible with ps6 only now.

kubernetes/configure
```diff
@@ -35,3 +42,48 @@ if [[ `juju status keystone --format=json 2>/dev/null| jq '.machines| length'` -
...
+    echo "done"
+done
+
+cat << EOF > $HOME/containerd-proxy.conf
+[Service]
+Environment="HTTP_PROXY=http://squid.internal:3128"
+Environment="HTTPS_PROXY=http://squid.internal:3128"
+Environment="NO_PROXY=127.0.0.1,10.0.0.0/8,192.168.0.0/16,172.16.0.0/12,registry.jujucharms.com,rocks.canonical.com,autocert.canonical.com,keystone.ps6.canonical.com,neutron.ps6.canonical.com,nova.ps6.canonical.com,glance.ps6.canonical.com,api.charmhub.io"
+EOF
+
```
dnegreira, line 66: I would add an echo here telling users why/how to register.
Something along the lines:

`We need to add the docker credentials to your cloud in order not to be rate limited. You can create your account on https://hub.docker.com/".`

nicolasbock (changes requested):

lathiat:
I have submitted an alternative PR which specifically takes care of configuring the contaienrd proxy, by using the containerd juju config options instead of the systemd override. I feel this method would mostly be preferable, with the exception that I don't auto-detect if we are in ProdStack6 or not:
https://github.com/canonical/stsstack-bundles/pull/218

For handling DockerHub authentication, there is also a charm option to set this which I think would also be preferred over the current method in this PR, unless I am missing something. I did not implement this in my PR. So perhaps if my PR gets merged, this one can be modified to handle only the Dockerhub authentication with some kind of convenience wrapper around setting the DockerHub username/password:
https://charmhub.io/containerd/configuration#custom_registries

sombrafam:
Dropping this one in favor of [Trents PR](https://github.com/canonical/stsstack-bundles/pull/218), that has a better solution.

PR #205, opened 2024-07-22
fix FIP allocation on prodstack6

dnegreira:
Since the allocations changed on prodstack6 to /25 we need to adapt the ./configure script to be more smart about which subnets we need to allocate to the FIP pool.

### DIFF
--- a/openstack/profiles/prodstack6
+++ b/openstack/profiles/prodstack6
@@ -3,7 +3,15 @@ source ~/novarc
 # this is currently assumed to be a /25 network
 EXT_SUBNET=subnet_${OS_USERNAME}-psd-extra
 CIDR=`openstack subnet show $EXT_SUBNET -c cidr -f value`
-ABC=${CIDR%.0*}
+# We reserve the last 64 of the /25 to FIP.
+# We get the last usable IP on the subnet
+FIP_RANGE_LASTIP=`openstack subnet show subnet_stg-reproducer-dnegreira-psd-extra -c allocation_pools -f json | jq -r '.allocation_pools[0].end'`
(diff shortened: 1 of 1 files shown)


openstack/profiles/prodstack6
```diff
@@ -3,7 +3,15 @@ source ~/novarc
 # this is currently assumed to be a /25 network
 EXT_SUBNET=subnet_${OS_USERNAME}-psd-extra
 CIDR=`openstack subnet show $EXT_SUBNET -c cidr -f value`
-ABC=${CIDR%.0*}
+# We reserve the last 64 of the /25 to FIP.
+# We get the last usable IP on the subnet
+FIP_RANGE_LASTIP=`openstack subnet show subnet_stg-reproducer-dnegreira-psd-extra -c allocation_pools -f json | jq -r '.allocation_pools[0].end'`
```
drencrom, line 8: This should be `$EXT_SUBNET` instead of `subnet_stg-reproducer-dnegreira-psd-extra`

drencrom (changes requested):

nicolasbock (changes requested):

openstack/profiles/prodstack6
dnegreira: Ops :)

dnegreira: fixed on https://github.com/canonical/stsstack-bundles/pull/205/commits/bd4bee6de926f751b33d33c387b18ec8b7de627a by @nicolasbock

nicolasbock (approved):
Thanks @dnegreira !

PR #209, opened 2024-07-26
Replace Sh shell in shebang with Bash

wilkmar:
These two scripts still used Sh shell in shebang. In some situations, it was an issue when they sourced Bash specific scripts (directly or indirectly)

See PR #208 for details


pponnuvel (approved):
LGTM

nicolasbock (approved):

PR #212, opened 2024-07-31
Fix Octavia upload image script

nicolasbock:
The retrofit action needs an image ID.

### DIFF
--- a/openstack/tools/upload_octavia_amphora_image.sh
+++ b/openstack/tools/upload_octavia_amphora_image.sh
@@ -1,2 +1,14 @@
 #!/bin/bash -eu
-juju run octavia-diskimage-retrofit/0 retrofit-image
+
+source $(realpath $(dirname $0))/../novarc
+
+IMAGE_ID=$(openstack image list --name jammy --format value --column ID)
+
+if [[ -n ${IMAGE_ID} ]]; then
+    juju run octavia-diskimage-retrofit/0 retrofit-image source-image=${IMAGE_ID}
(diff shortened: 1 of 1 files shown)


openstack/tools/upload_octavia_amphora_image.sh
```diff
@@ -1,2 +1,14 @@
 #!/bin/bash -eu
-juju run octavia-diskimage-retrofit/0 retrofit-image
+
+source $(realpath $(dirname $0))/../novarc
+
+IMAGE_ID=$(openstack image list --name jammy --format value --column ID)
```
brianphaley, line 5: So this script previously took a release name via '--release foo' which is useful when running on a focal deploy. Defaulting to jammy won't work there so perhaps it should be put back, unless we can determine the release from something else?

brianphaley (changes requested):

openstack/tools/upload_octavia_amphora_image.sh
nicolasbock: Ok, that's a good point. I'll work on this PR a bit more.

dosaboy: I believe this is not necessary if we have https://github.com/canonical/stsstack-bundles/pull/217 i.e. of an image exists in glance that corresponds to the version/release deployed (focal->ussuri, jammy->yoga etc).

nicolasbock:
This is not needed, see #291

PR #213, opened 2024-08-01
Fix prodstack6 VIP allocation

arif-ali:
At the moment the VIPs start from 106, even though the ip range may start from 129.

We use `ipcalc` command to find the HostMin, and calculate cased on that

### DIFF
--- a/common/generate_bundle_base
+++ b/common/generate_bundle_base
@@ -23,7 +23,9 @@ if [[ -z $vip_start ]] && [[ -e ~/novarc ]]; then
         # Our primary network is /25 and we are arbitrarily using the
         # last 20 addresses for vips which is prone to collisions but
         # we have no alternative currently.
-        vip_start=$(echo $cidr| sed -r 's/([0-9]+\.[0-9]+\.[0-9]+).+/\1/g').106
+        net_start=$(awk -F'.' '/HostMin/{print $NF}' <<<$(ipcalc -b $cidr))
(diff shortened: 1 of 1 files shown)


common/generate_bundle_base
```diff
@@ -23,7 +23,9 @@ if [[ -z $vip_start ]] && [[ -e ~/novarc ]]; then
         # Our primary network is /25 and we are arbitrarily using the
         # last 20 addresses for vips which is prone to collisions but
         # we have no alternative currently.
-        vip_start=$(echo $cidr| sed -r 's/([0-9]+\.[0-9]+\.[0-9]+).+/\1/g').106
+        net_start=$(ipcalc -b $cidr |grep HostMin | awk '{print $2}')
```
pponnuvel, line 26: You can avoid the extra pipe with: `net_start=$(ipcalc -b $cidr | awk '/HostMin/{print $2}')`

Alternatively, can also simplify these lines with (which avoids pipes altogether):
`net_start_start=$(awk -F'.' '/HostMin/{print $NF}' <<<$(ipcalc -b $cidr))`

pponnuvel (approved):
A minor suggestion but looks fine regardless.

common/generate_bundle_base
arif-ali: nice, I like that, will amend

dnegreira (approved):

PR #215, opened 2024-08-02
Add a script to allocate VIPs in PS6

arif-ali:
This eneusres that these IP addresses don't get used by OpenStack and the relevant VIPs can be utilised and hence don't have conflicts

### DIFF
--- a/openstack/tools/allocate_vips.sh
+++ b/openstack/tools/allocate_vips.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# Allocate the last 20 IP addresses, which could be considered for
+# VIPs in OpenStack
+
+source ~/novarc
+num_vips=20
+subnet="subnet_${OS_USERNAME}-psd"
+
+cidr=$(openstack subnet show ${subnet} -c cidr -f value 2>/dev/null)
+
+net_end=$(awk -F'.' '/HostMax/{print $NF}' <<<$(ipcalc -b $cidr))
+
+vip_start_suffix=$((net_end - num_vips + 1))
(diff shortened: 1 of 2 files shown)


openstack/tools/allocate_vips.sh
```diff
@@ -0,0 +1,27 @@
...
+
+source ~/novarc
+num_vips=20
+subnet="subnet_${OS_USERNAME}-psd"
+
+cidr=$(openstack subnet show ${subnet} -c cidr -f value 2>/dev/null)
+
+net_end=$(awk -F'.' '/HostMax/{print $NF}' <<<$(ipcalc -b $cidr))
+
+net_end_suffix=${net_end##*\.}
```
pponnuvel, line 14: This line isn't needed at all. It was already handled by awk.

Likewise you can get rid of [this line](https://github.com/canonical/stsstack-bundles/pull/213/files#diff-24e887a57e2041b3f5464aec5b017164cb68e5d0b70939d342390e6eb276302cR27) from `common/generate_bundle_base`.

openstack/tools/allocate_vips.sh
```diff
@@ -0,0 +1,27 @@
...
+
+vip_start_suffix=$((net_end_suffix - ${num_vips} + 1))
```
pponnuvel, line 16: Parameter expansion is performed in arithmetic expansion too, so you can drop the PE: `vip_start_suffix=$((net_end_suffix - num_vips + 1))`

pponnuvel (approved):
LGTM

dnegreira (approved):

PR #217, opened 2024-08-13
Add focal image to glance for openstack

dosaboy:
This is to allow octavia-diskimage-retrofit to work without having to provide any args our source-image id since it will look for the image corresponding to the release of openstack deployed (using image
properties).


brianphaley (approved):
lgtm since both images always seem to be there

PR #218, opened 2024-08-14
kubernetes: Configure proxy for containerd

lathiat:
Configure the containerd proxy server, as it's not possible to pull from
most image repositories without it in se-cloud.

Defaults to the required settings for the se-cloud environment. Can be
overridden with --containerd-proxy and --containerd-no-proxy.

For those using stsstack-bundles outside of the se-cloud environment,
use --containerd-use-proxy false to skip it. This is required as passing
a blank ("") argument to --containerd-proxy is not parsed correctly.


kubernetes/module_defaults
```diff
@@ -17,4 +17,5 @@ MOD_PARAMS[__NUM_K8S_WORKER_UNITS__]=2
 MOD_PARAMS[__NUM_K8S_LB_UNITS__]=1
 MOD_PARAMS[__ETCD_SNAP_CHANNEL__]='latest/stable'
 MOD_PARAMS[__CONTAINER_RUNTIME__]='containerd'
-
+MOD_PARAMS[__CONTAINERD_PROXY__]='http://squid.internal:3128'
```
dosaboy, line 20: rather than setting proxy by default lets leave empty and populate if we are running in prodstack which we can detect by doing e.g. ```if `juju show-cloud prodstack &>/dev/null`; then ...```

kubernetes/pipeline/02configure
```diff
@@ -110,6 +110,22 @@ do
...
+        --containerd-proxy)  #__OPT__type:<str> (default=http://squid.internal:3128)
+	    echo "XXX X: ${MOD_PARAMS[__CONTAINERD_PROXY__]} Y: $2"
+            MOD_PARAMS[__CONTAINERD_PROXY__]=$2
+            shift
+            ;;
+        --containerd-no-proxy)  #__OPT__type:<str> (default=127.0.0.1,localhost,::1,10.149.0.0/16,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16)
+            MOD_PARAMS[__CONTAINERD_NO_PROXY__]=$2
+            shift
+            ;;
         --calico)
```
sombrafam, line 129: We should also add options, for users to configure the docker user and password. The following config worked for me:

```
juju config containerd custom_registries='[{"url": "registry-1.docker.io", "username": "sombrafam", "password": "********"}]'
```
We should also print a warning if the user does not set those, since newer users will likely not know that this is almost mandatory and it's hard to diagnose once the deployment fails.

sombrafam (changes requested):

kubernetes/module_defaults
lathiat: Implemented this

microk8s/module_defaults
```diff
@@ -8,3 +8,5 @@
 
 MOD_PARAMS[__MICROK8S_CHANNEL__]="latest/edge"
 MOD_PARAMS[__NUM_MICROK8S_UNITS__]=1
+MOD_PARAMS[__CONTAINERD_PROXY__]='http://squid.internal:3128'
+MOD_PARAMS[__CONTAINERD_NO_PROXY__]='127.0.0.1,localhost,::1,10.149.0.0/16,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16'
```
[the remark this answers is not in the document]
lathiat: I got these added to the proxy rules instead so it's not required

kubernetes/pipeline/02configure
lathiat: I did a bunch of deployments without hitting these limits.. any idea what you need to deploy how many times to hit this?

lathiat: Think I'd prefer to address this in a separate commit to get this main one in first

lathiat:
Updated PR to auto-set proxy based on whether the hostname squid.internal resolves. Intentionally left the DockerHub credentials out for now, will look to address that in a separate PR.

Ready for review/merge.

dosaboy (approved):

kubernetes/pipeline/02configure
sombrafam: Maybe it's related to which of the pain you're connected to? I hit these many times.

PR #220, opened 2024-08-28
Add openstack charm func test runner

dosaboy:
Also includes tool to identify func test targets from the charm.

### DIFF
--- a/openstack/tools/charmed_openstack_functest_runner.sh
+++ b/openstack/tools/charmed_openstack_functest_runner.sh
@@ -1,0 +1,159 @@
+#!/bin/bash -eu
+#
+# Run Charmed Openstack CI tests manually in a similar way to how they are run
+# by OpenStack CI (OSCI).
+#
+# Usage: clone/fetch charm to test and run from within charm root dir.
+#
+FUNC_TEST_PR=
+FUNC_TEST_TARGET=
+MODIFY_BUNDLE_CONSTRAINTS=true
+SKIP_BUILD=false
+
+usage () {
+    cat << EOF
+USAGE: `basename $0` OPTIONS
+
+Run OpenStack charms functional tests manually in a similar way to how
+Openstack CI (OSCI) would do it. This tool should be run from within a charm
+root.
+
+Not all charms use the same versions and dependencies and an attempt is made to
+cover this here but in some cases needs to be dealt with as a pre-requisite to
+running the tool. For example some charms need their tests to be run using
+python 3.8 and others python 3.10. Some tests might require Juju 2.9 and others
+Juju 3.x - the assumption in this runner is that Juju 3.x is ok to use.
+
+OPTIONS:
+    --func-test-target
+        Provide the name of a specific test target to run. If none provided
+        all tests are run based on what is defined in osci.yaml i.e. will do
+        what osci would do by default.
+    --func-test-pr
+        Provides similar functionality to Func-Test-Pr in commit message. Set
+        to zaza-openstack-tests Pull Request ID.
+    --skip-build
+        Skip building charm if already done to save time.
+    --skip-modify-bundle-constraints
+        By default we modify test bundle constraints to ensure that applications
+        have the resources they need. For example nova-compute needs to have
+        enough capacity to boot the vms required by the tests.
+    --help
+        This help message.
+EOF
+}
+
+while (($# > 0))
+do
+    case "$1" in
+        --func-test-target)
+           FUNC_TEST_TARGET=$2
+           shift
+           ;;
+        --func-test-pr)
+           FUNC_TEST_PR=$2
+           shift
+           ;;
+        --skip-modify-bundle-constraints)
+          MODIFY_BUNDLE_CONSTRAINTS=false
+          ;;
+        --skip-build)
+          SKIP_BUILD=true
+          ;;
+        --help|-h)
+          usage
+          exit 0
+          ;;
+        *)
+          echo "ERROR: invalid input '$1'"
+          usage      
+          exit 1
+          ;;
+    esac
+    shift
+done
+
+set -x
+
+source ~/novarc
+export {,TEST_}CIDR_EXT=`openstack subnet show subnet_${OS_USERNAME}-psd-extra -c cidr -f value`
+FIP_MAX=$(ipcalc $CIDR_EXT| awk '$1=="HostMax:" {print $2}')
+FIP_MIN=$(ipcalc $CIDR_EXT| awk '$1=="HostMin:" {print $2}')
+FIP_MIN_ABC=${FIP_MIN%.*}
+FIP_MIN_D=${FIP_MIN##*.}
+FIP_MIN=${FIP_MIN_ABC}.$(($FIP_MIN_D + 64))
+
+CIDR_OAM=`openstack subnet show subnet_${OS_USERNAME}-psd -c cidr -f value`
+OAM_MAX=$(ipcalc $CIDR_OAM| awk '$1=="HostMax:" {print $2}')
+OAM_MIN=$(ipcalc $CIDR_OAM| awk '$1=="HostMin:" {print $2}')
+OAM_MIN_ABC=${OAM_MIN%.*}
+OAM_MAX_D=${OAM_MAX##*.}
+# Picking last two addresses and hoping they dont get used by Neutron.
+export OS_VIP00=${OAM_MIN_ABC}.$(($OAM_MAX_D - 1))
+export OS_VIP01=${OAM_MIN_ABC}.$(($OAM_MAX_D - 2))
+
+# More information on config https://github.com/openstack-charmers/zaza/blob/master/doc/source/runningcharmtests.rst
+export {,TEST_}NET_ID=$(openstack network show net_${OS_USERNAME}-psd-extra -f value -c id)
+export {,TEST_}FIP_RANGE=$FIP_MIN:$FIP_MAX
+export {,TEST_}GATEWAY=$(openstack subnet show subnet_${OS_USERNAME}-psd-extra -c gateway_ip -f value)
+export {,TEST_}NAME_SERVER=91.189.91.131
+export {,TEST_}CIDR_PRIV=192.168.21.0/24
+#export SWIFT_IP=10.140.56.22
+export TEST_MODEL_SETTINGS="image-stream=released;default-series=jammy;test-mode=true;transmit-vendor-metrics=false"
+# We need to set TEST_JUJU3 as well as the constraints file
+# Ref: https://github.com/openstack-charmers/zaza/blob/e96ab098f00951079fccb34bc38d4ae6ebb38606/setup.py#L47
+export TEST_JUJU3=1
+
+# NOTE: this should not be necessary for > juju 2.x but since we still have a need for it we add it in
+export TEST_ZAZA_BUG_LP1987332=1
+
+# Some charms point to an upstream constraints file that installs python-libjuju 2.x so we need to do this to ensure we get 3.x
+export TEST_CONSTRAINTS_FILE=https://raw.githubusercontent.com/openstack-charmers/zaza/master/constraints-juju34.txt
+
+# 2. Build
+if ! $SKIP_BUILD; then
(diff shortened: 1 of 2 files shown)


openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash -ux
+source ~/novarc
+CIDR=`openstack subnet show subnet_${OS_USERNAME}-psd-extra -c cidr -f value`
+FIP_MAX=$(ipcalc $CIDR| awk '$1=="HostMax:" {print $2}')
+FIP_MIN=$(ipcalc $CIDR| awk '$1=="HostMin:" {print $2}')
+FIP_MIN_ABC=${FIP_MIN%.*}
+FIP_MIN_D=${FIP_MIN##*.}
+FIP_MIN=${FIP_MIN_ABC}.$(($FIP_MIN_D + 64))
```
brianphaley, line 8: This turns out to be gateway_ip + 64 since it's always .1 by default, and as long as this is a /25 or larger it's fine, smaller and it will be greater than the max. It's only until they get the gate fixed anyways...

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -0,0 +1,64 @@
...
+export TEST_GATEWAY=$(openstack subnet show subnet_${OS_USERNAME}-psd-extra -c gateway_ip -f value)
+export TEST_NAME_SERVER=91.189.91.131
+export TEST_CIDR_PRIV=192.168.21.0/24
+export TEST_SWIFT_IP=10.140.56.22
+export TEST_MODEL_SETTINGS="image-stream=released;default-series=jammy;test-mode=true;transmit-vendor-metrics=false"
+export TEST_CONSTRAINTS_FILE=https://raw.githubusercontent.com/openstack-charmers/zaza/master/constraints-juju34.txt
+# We need to set TEST_JUJU3 as well as the constraints file
+# Ref: https://github.com/openstack-charmers/zaza/blob/e96ab098f00951079fccb34bc38d4ae6ebb38606/setup.py#L47
+export TEST_JUJU3=1
+export NET_ID=$(openstack network show net_${OS_USERNAME}-psd -f value -c id)
```
brianphaley, line 23: Just making sure this wasn't supposed to be -psd-extra network

xtrusia:
While Im testing mysql-router charm zaza test, I realized that I need below env variable

export TEST_ZAZA_BUG_LP1987332=1

related code is here

https://github.com/openstack-charmers/zaza/blob/bef1bdbac7ee15ae742466641c65bcab363c093e/zaza/utilities/juju.py#L511

if I didn't set it, there was python array index error.

just FYI

xtrusia:
Is there way to use PR test ? like below in commit msg

func-test-pr: https://github.com/openstack-charmers/zaza-openstack-tests/pull/1252

dosaboy:
> While Im testing mysql-router charm zaza test, I realized that I need below env variable
> 
> export TEST_ZAZA_BUG_LP1987332=1
> 
> related code is here
> 
> https://github.com/openstack-charmers/zaza/blob/bef1bdbac7ee15ae742466641c65bcab363c093e/zaza/utilities/juju.py#L511
> 
> if I didn't set it, there was python array index error.
> 
> just FYI

I beleive this is not needed with juju3 but can add it anyway for those wanted to stick with j2

dosaboy:
> Is there way to use PR test ? like below in commit msg
> 
> func-test-pr: [openstack-charmers/zaza-openstack-tests#1252](https://github.com/openstack-charmers/zaza-openstack-tests/pull/1252)

I have a need for this myself so would like to find a way but do not yet know how

openstack/tools/charmed_openstack_functest_runner.sh
dosaboy: ah well spotted, this is the "external" network used for fips so for us thats \*-psd-extra

dosaboy:
With the current version I can now do a successful build and deploy but hit the following error once the tests start:
```
2024-08-30 11:29:25 [ERROR] Model default_alias (zaza-e680c75cf4a9)
Traceback (most recent call last):
  File "/home/ubuntu/jammy/octavia/src/.tox/func-target/bin/functest-run-suite", line 8, in <module>
    sys.exit(main())
  File "/home/ubuntu/jammy/octavia/src/.tox/func-target/lib/python3.10/site-packages/zaza/charm_lifecycle/func_test_runner.py", line 405, in main
    func_test_runner(
  File "/home/ubuntu/jammy/octavia/src/.tox/func-target/lib/python3.10/site-packages/zaza/charm_lifecycle/func_test_runner.py", line 313, in func_test_runner
    run_env_deployment(env_deployment, keep_model=preserve_model,
  File "/home/ubuntu/jammy/octavia/src/.tox/func-target/lib/python3.10/site-packages/zaza/charm_lifecycle/func_test_runner.py", line 174, in run_env_deployment
    configure.configure(
  File "/home/ubuntu/jammy/octavia/src/.tox/func-target/lib/python3.10/site-packages/zaza/charm_lifecycle/configure.py", line 58, in configure
    run_configure_list(functions)
  File "/usr/lib/python3.10/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/home/ubuntu/jammy/octavia/src/.tox/func-target/lib/python3.10/site-packages/zaza/charm_lifecycle/configure.py", line 44, in run_configure_list
    utils.get_class(func)()
  File "/home/ubuntu/jammy/octavia/src/.tox/func-target/lib/python3.10/site-packages/zaza/openstack/charm_tests/neutron/setup.py", line 157, in basic_overcloud_network
    undercloud_and_charm_setup(limit_gws=limit_gws)
  File "/home/ubuntu/jammy/octavia/src/.tox/func-target/lib/python3.10/site-packages/zaza/openstack/charm_tests/neutron/setup.py", line 107, in undercloud_and_charm_setup
    network.setup_gateway_ext_port(network_config,
  File "/home/ubuntu/jammy/octavia/src/.tox/func-target/lib/python3.10/site-packages/zaza/openstack/configure/network.py", line 315, in setup_gateway_ext_port
    openstack_utils.configure_gateway_ext_port(
  File "/home/ubuntu/jammy/octavia/src/.tox/func-target/lib/python3.10/site-packages/zaza/openstack/utilities/openstack.py", line 1125, in configure_gateway_ext_port
    net_id = get_admin_net(neutronclient)['id']
TypeError: 'NoneType' object is not subscriptable
```

dosaboy:
I have hopefully fixed the above issue which might have been caused by the fact that had forgotten to export CIDR_EXT but I also notice that both the zaza and zaza-openstack-tests repos have duplicate code for e.g. https://github.com/openstack-charmers/zaza/blob/master/zaza/utilities/generic.py#L92 https://github.com/openstack-charmers/zaza-openstack-tests/blob/master/zaza/openstack/utilities/generic.py#L153 which are slightly different in which env vars they look for and it is not clear which is actually used when a test is run so I am now setting both kinds.

dosaboy:
works now but new issue is that nova-compute vms use a flavor too small to be able to boot any vms

dosaboy:
> works now but new issue is that nova-compute vms use a flavor too small to be able to boot any vms

```
2024-08-30 13:40:04.351 230827 INFO nova.scheduler.manager [None req-f758d68f-0f19-461c-9068-cb4e96e272e7 184d6a2c973043d8b5ac655217b4c3cf ac3a5e14642c4a9dad8fcde1b0c3530e - - c8b548e494dd4827aa0418706da3f6a7 c8b548e494dd4827aa0418706da3f6a7] Got no allocation candidates from the Placement API. This could be due to insufficient resources or a temporary occurrence as compute nodes start up.
```

dosaboy:
Putting this up for final review since I am now able to successfully run tests.

rodrigogansobarbieri:
I tried running the script but I got the same error:

Juju.errors.JujuConnectionError: juju server-version 3.5.3 not supported

I changed my juju snap version to 2.9 and pointed the credentials to my juju 2.9 controller and it worked. It is not running the tests. I guess whether it supports juju 3 or not varies from charm to charm. I am running nova-compute charm tests.

One thing I noticed is that the timeout is set to 3600 instead of 7200 as Trent suggested in his notes: export TEST_DEPLOY_TIMEOUT=7200

Also one thing I don't like is that there is currently to option to specify the bundle we want to run. For example, for my nova-compute tests I am running for the 2023.2 branch, so it started running jammy-antelope, and then later it will run jammy-bobcat. Assuming jammy-bobcat fails for some reason and I have to run it again. If I have already collected a successful output from jammy-antelope I would not want to let it run again, but instead specify that I want only to run jammy-bobcat.

rodrigogansobarbieri (changes requested):
-1 for lacking an option to specify the bundle to run (example: run just jammy-bobcat instead of the whole suite that would include jammy-antelope, jammy-bobcat and so on).

dosaboy:
> -1 for lacking an option to specify the bundle to run (example: run just jammy-bobcat instead of the whole suite that would include jammy-antelope, jammy-bobcat and so on).

done.

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -0,0 +1,121 @@
...
+#
+# Usage: clone/fetch charm to test and run from within charm root dir.
+#
+FUNC_TEST_PR=
+FUNC_TEST_TARGET=
+SKIP_BUILD=false
+
+usage () { echo "USAGE: --func-test-target | --func-test-pr | --skip-build"; }
+
+while (($# > 0))
```
lathiat, line 14: If you call the script without any arguments and you're not inside a charm directory it eventually fails. It should check for the current directory being a charm somehow.

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -0,0 +1,121 @@
...
+# Run Charmed Openstack CI tests manually in a similar way to how they are run
+# by OpenStack CI (OSCI).
+#
+# Usage: clone/fetch charm to test and run from within charm root dir.
+#
+FUNC_TEST_PR=
+FUNC_TEST_TARGET=
+SKIP_BUILD=false
+
+usage () { echo "USAGE: --func-test-target | --func-test-pr | --skip-build"; }
```
lathiat, line 12: Ideally this would make it more clear how to use the tool, e.g. that you have to call it from within a charm directory, etc.

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -0,0 +1,121 @@
...
+fi
+
+for target in ${func_targets[@]}; do
+    [[ -d src ]] && pushd src || true
+    tox -re func-target -- $target
+
+    read -p "Destroy model and run next test? [ENTER]"
+    # cleanup before next run
+    model=`juju list-models| egrep -o "^zaza-\S+"|tr -d '*'`
+    juju destroy-model --no-prompt $model --force --no-wait --destroy-storage
```
lathiat, line 119: This will fail if you have multiple matching models. Ideally it would also print the name of all the models it will destroy before it does, in case you had created another model with a name starting with zaza.

You can also just tell zaza to destroy the model itself automatically by passing an argument, perhaps we should use that by default and then have an option to switch to manual cleanup behaviour for trying to debug the tests?

lathiat (changes requested):
These are nits, and not really related to the tool's core functionality. But thought they are worth adding/fixing.

lathiat:
Tested working OK for me with glance-simplestreams-sync yoga

openstack/tools/charmed_openstack_functest_runner.sh
dosaboy: do you happen to know what the argument is?

dosaboy: we can fix this as a future patch

openstack/tools/charmed_openstack_functest_runner.sh
dosaboy: i have put in a much nicer help message

openstack/tools/charmed_openstack_functest_runner.sh
dosaboy: i will fix this as a followup patch but have put info in the help message about how to run the tool

PR #222, opened 2024-09-04
Force Juju to use Ceph for vm root disk

dosaboy:
Adds a --model-constraints flag to override the default which is now to use a cinder/ceph volume for root disk.


common/helpers
```diff
@@ -181,6 +182,14 @@ filter_master_opts ()
                 fi
                 shift
                 ;;
+            --model-constraints)
+                if [ -z "${MASTER_OPTS[MODEL_CONSTRAINTS]}" ]; then
```
pponnuvel, line 186: Think you can just append unconditionally.

pponnuvel (approved):
LGTM

dosaboy:
quotas have been updated to support this increase in volume usage so we're good to go now

Issue #227, opened 2024-09-06
enable stsstack to also support non-openstack clouds


zhhuabj:
In fact, stsstack can also work well on lxd cloud. so can we enable stsstack to not only support underlying openstack cloud, but also support underlying lxd and maas cloud as well ?

PR #228, opened 2024-09-06
enable stsstack to also support non-openstack clouds #227

zhhuabj:
In fact, stsstack can work well on lxd cloud as well. So this patch is to enable stsstack to not only support underlying openstack cloud, but also support underlying lxd and maas cloud.


common/generate_bundle_base
```diff
@@ -9,13 +9,19 @@ declare -a overlay_opts=()
 # Catch any changes/additions to master opts
 update_master_opts ${MOD_PASSTHROUGH_OPTS[@]}
 
+type=$(get_cloud_type)
 vip_start=${MASTER_OPTS[VIP_ADDR_START]}
+cidr=""
 if [[ -z $vip_start ]] && [[ -e ~/novarc ]]; then
```
pponnuvel, line 15: Think you can simply change this by changing:
```
if [[ -z $vip_start ]] && [[ -e ~/novarc ]]; then
```

to:
```
if [[ -z $vip_start ]] && [[ -e ~/novarc ]] && [[ $type = openstack ]]; then
```

Then you wouldn't check twice inside the if. And can also drop the `cidr=""` too.

common/helpers
```diff
@@ -43,6 +43,13 @@ list_overlays ()
     } | sort -u
 }
 
+get_cloud_type ()
+{
+   local cloud=`juju show-model| sed -rn 's/.+cloud:\s*(.+).*/\1/p'| uniq`
```
pponnuvel, line 48: `jq` is better than parsing yaml with sed (it can also get unique vales):
```
   local cloud=$(juju show-model --format json |jq -r '[.[].cloud] | unique[]')
   local type=$(juju show-cloud $cloud --format json | jq -r '[.[].type] | unique[]')
```

```

File: common/generate_bundle_base [hunk 3]
```diff
@@ -9,13 +9,19 @@ declare -a overlay_opts=()
...
     if [[ -z $cidr ]]; then
-        # stsstack
-        cidr=$(source ~/novarc; openstack subnet show ${OS_USERNAME}_admin_subnet -c cidr -f value 2>/dev/null)
+        if [[ $type = openstack ]]; then
+            # stsstack
+            cidr=$(source ~/novarc; openstack subnet show ${OS_USERNAME}_admin_subnet -c cidr -f value 2>/dev/null)
+        fi
         if [[ -n $cidr ]]; then
             vip_start=$(echo $cidr| sed -r 's/([0-9]+\.[0-9]+).+/\1/g').150.0
         fi
```
brianphaley, line 27: I think all the above code should be under the 'openstack' if since it is only used in that case, correct? Then you wouldn't need the cidr="" assignment either.

brianphaley (changes requested):

common/generate_bundle_base
zhhuabj: hi brian, I have added the following code for non-openstack cloud to improve this scene. thanks for the review.

elif [[ -z $vip_start  ]]; then
    echo "ERROR: No VIP start address provided with --vip-addr-start for the $type cloud."
exit 1

common/generate_bundle_base
zhhuabj: hi pon, I made the modifications as you suggested, 

if [[ -z $vip_start ]] && [[ -e ~/novarc ]] && [[ $type = openstack ]]; then

and also added the following code to improve it. thanks.

elif [[ -z $vip_start  ]]; then
    echo "ERROR: No VIP start address provided with --vip-addr-start for the $type cloud."
exit 1

common/helpers
zhhuabj: Hi pon, yes, jq is better than parsing yaml with sed. Although I merely relocated two original lines without modifying the previous code (perhaps the optimized code should submit another pr), I have now modified to using jq instead of sed to align with your suggestion. thanks for the review.

pponnuvel (approved):
LGTM

common/helpers
pponnuvel: Thanks Joshua. Looks good now!

dosaboy (approved):

PR #235, opened 2024-09-11
charm test runner: manual_functest and remote-build

rodrigogansobarbieri:
- Add support for remote-build for when the tests need to be run from a container due to python version
- Add support for manual functest runners (deploy, configure, test) for when the full suite fails requiring more granular approach (such as a sleep)

### DIFF
--- a/openstack/tools/charmed_openstack_functest_runner.sh
+++ b/openstack/tools/charmed_openstack_functest_runner.sh
@@ -40,7 +40,7 @@ OPTIONS:
         By default we wait before destroying the model after a test run. This
         flag can used to override that behaviour.
     --manual-functests
-        Runs functest commands separately (deploy,configure,test)instead of
+        Runs functest commands separately (deploy,configure,test) instead of
         the entire suite.
     --remote-build
         Builds the charm in a remote location and transfers the charm file over.
--- a/openstack/tools/charmed_openstack_functest_runner.sh
+++ b/openstack/tools/charmed_openstack_functest_runner.sh
@@ -7,8 +7,11 @@
 #
 FUNC_TEST_PR=
 FUNC_TEST_TARGET=
+MANUAL_FUNCTESTS=false
 MODIFY_BUNDLE_CONSTRAINTS=true
+REMOTE_BUILD=
 SKIP_BUILD=false
+SLEEP=
 WAIT_ON_DESTROY=true
 
 usage () {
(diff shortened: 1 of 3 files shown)


openstack/tools/func_test_tools/manual_functests_runner.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash -eu
```
dosaboy, line 1: can you add this file to tox.ini so that it gets tested by bashate

dosaboy: a docstring might be good to explain why and when this is useful to use

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -36,12 +39,22 @@ OPTIONS:
     --no-wait
         By default we wait before destroying the model after a test run. This
         flag can used to override that behaviour.
+    --manual-functests
+        Runs functest commands separately (deploy,configure,test)instead of
```
dosaboy, line 43: missing space before "instead"

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -36,12 +39,22 @@ OPTIONS:
...
+        the entire suite.
+    --remote-build
+        Builds the charm in a remote location and transfers the charm file over.
+        The destination needs to be prepared for the build and authorized for
+        ssh. Implies --skip-build. Specify parameter as <destination>,<path>.
+        Example: --remote-build ubuntu@10.171.168.1,~/git/charm-nova-compute
```
dosaboy, line 49: do we really need this? cant you do the build on vm then rsync it all to a container and run the tests?

dosaboy: otherwise we are just adding more and more complexity when there are alternatives

openstack/tools/func_test_tools/manual_functests_runner.sh
rodrigogansobarbieri: thanks, will do

openstack/tools/charmed_openstack_functest_runner.sh
rodrigogansobarbieri: thanks, will do.

openstack/tools/charmed_openstack_functest_runner.sh
rodrigogansobarbieri: yes, we need this if we are going to script things to avoid human mistakes as much as possible and have everything performed by the script. If things are done manually someone could build the wrong version, rsync the wrong file, etc, and the resulting script that does not include those steps wouldn't capture the mistake.

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -139,6 +164,15 @@ if ! $SKIP_BUILD; then
     lxd init --auto || true
 
     tox -re build
+else
```
dosaboy, line 167: this could be an elif

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -36,12 +39,22 @@ OPTIONS:
     --no-wait
         By default we wait before destroying the model after a test run. This
         flag can used to override that behaviour.
+    --manual-functests
+        Runs functest commands separately (deploy,configure,test) instead of
+        the entire suite.
+    --remote-build
```
dosaboy, line 45: since this takes arguments you could specify that in the help message e.g.

```
--remote-build USER@HOST,GIT_PATH
```

dosaboy (approved):
happy to fix the nits i pointed out as a followup patch

Issue #234, opened 2024-09-11
charm func test runner: vault initialization is not running


xtrusia:
While I testing jammy-yoga-vault,  I saw vault is not initialized.

related info is in tests.yaml

for now, I'm not sure how to solve it. but I'll analyze it.

thanks

https://review.opendev.org/c/openstack/charm-manila-ganesha/+/925662/4/src/tests/tests.yaml

xtrusia:
#238 solved this issue

PR #236, opened 2024-09-12
Taking additional targets from osci.yaml

xtrusia:
osci also has targets so added them.

[1] https://review.opendev.org/c/openstack/charm-manila-ganesha/+/925662/5/osci.yaml

Resolves: #234 

currently it runs 
focal-yoga
jammy-yoga from tests.yaml

but I think vault:jammy-yoga-vault also needed from osci.yaml

image

### DIFF
--- a/openstack/tools/func_test_tools/identify_charm_func_tests.py
+++ b/openstack/tools/func_test_tools/identify_charm_func_tests.py
@@ -8,6 +8,7 @@ import os
 
 import yaml
 
+OSCI_YAML = 'osci.yaml'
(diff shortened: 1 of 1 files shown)


xtrusia:
I need to check duplicated job between tests.yaml and osci.yaml

dosaboy:
having discussed with the Openstack team how this works and why the manila-ganehsa charm has the job configured that way i wrote https://github.com/canonical/stsstack-bundles/pull/238/files which can supersede this patch.

dosaboy:
replaced by 238

PR #238, opened 2024-09-12
Add support for aliased zaza test targets

dosaboy:
Zaza supports defining aliased configure sections such that a test can be run under an alias by prefixig the target name with <alias>: which will result in zaza using the alias configure section rather than the default. This adds support for extrating these aliased targets from oscsi.yaml.

### DIFF
--- a/openstack/tools/func_test_tools/common.py
+++ b/openstack/tools/func_test_tools/common.py
@@ -1,0 +1,33 @@
+""" Common helpers for func test runners. """
+import os
+
+import yaml
+
+
+class OSCIConfig():
+    """ Extract information from osci.yaml """
+    def __init__(self):
+        if not os.path.exists('osci.yaml'):
+            self._osci_config = {}
+        else:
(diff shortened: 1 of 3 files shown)


xtrusia:
LGTM +1

Tested shortly 
1. run test without change
2. run test only with (aliased_bundles)
Didn't run all but checked functest-run-suite --keep-model --bundle vault:jammy-yoga-vault ( this part )

xtrusia (approved):

dosaboy:
From my test I can confirm that this did deploy the test env and unseal vault but I did hit a race condition where the test tries to ru before vault has finished executing all relations and the endpoints have been updated. This is why osci waits for all units to be idle before continuing.

I was however able to run the tests by doing:
```
cd src
. .tox/func-target/bin/activate
functest-test -m zaza-f5ff50b533f2
```

Maybe we could suggest that as an alternative if the tests fail.

PR #239, opened 2024-09-12
Ensure nova-compute constraints set

dosaboy:
Some charms do not set constraints at all for
nova-compute so this adjusts the code to ensure
they are set even if no constraints exist.


openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -161,10 +161,18 @@ else
...
 
+# Ensure nova-compute has enough resources to create vms in tests. Not all
+# charms have bundles with constraints set so we need to cover both cases here.
 if $MODIFY_BUNDLE_CONSTRAINTS; then
     (
     [[ -d src ]] && cd src
-    sed -i -r '/\s+nova-compute:$/{n;s/mem=[0-9]+M/root-disk=80G mem=8G/}' tests/bundles/*.yaml
+    for f in tests/bundles/*.yaml; do
+        before=$(md5sum $f)
+        sed -i -r '/\s+nova-compute:$/{n;s/mem=[0-9]+M/root-disk=80G mem=8G/}' $f
```
nicolasbock, line 171: This only matches if the unit is `M`. If the bundles uses `G` this would not work.

pponnuvel: If you can install `yq`, you can do:
```
yq -i '.applications.nova-compute.constraints = "root-disk=80G mem=8G"' $i
```
which would parse the yaml & work correctly and not as ugly or error-prone as `sed`.

`yq` isn't installed by default. But it's very useful and I don't see why we can't install it (as part of the stsstack-bundles setup).

dosaboy: done.

pponnuvel (approved):
LGTM

Issue #229, opened 2024-09-08
The default 80Gb for nova-compute's vda is a bit big


lccnblack:
The default size of vda as root disk is 80Gb, and it seems that we can change it by adjusting the parameter MOD_PARAMS[__NOVA_COMPUTE_UNIT_CONSTRAINTS __] (/stsstack-bundles/openstack/pipeline/02configure)to change the size 

The default vda is 30GB for normal nodes, suppose an OpenStack without HA, with ceph, Octavia and 2 nova-compute, 
Then the vda alone occupies no less than 600GB（20*30Gb), if we add 2*50Gb vdb, then it is very easy to meet the upper limit of 1000GB. 
for example
` ./generate-bundle.sh --no-ml2dns --vault --ceph --octavia --num-compute 2 --series focal
`

And in addition to having an OpenStack environment, we usually have other test environments in parallel.

And vdb already has 50Gb(/stsstack-bundles/openstack/b/o/openstack/vault-openstack-secrets.yaml) as an ephemeral disk directory (/var/lib/nova/instances), so why does vda need to have 80Gb of space?
BTW, 50Gb for nova-compute vdb, is it also too big?

I think 30Gb is more reasonable for root disk of nova-compute.

lathiat:
I independently came to the same conclusion, then found this existing bug, thanks @lccnblack 

The reason the vault overlay has a separate ephemeral disk is to exercise disk-encryption with vault. This made sense when vault wasn't the default, but since we added OVN (a long time ago now) this always gets pulled in by default.

I think a simple fix here is to revert nova-compute back to a 30G default root-disk, and always enable the 50G ephemeral disk even without vault.

I'm just testing the practical effect of this before I'll raise a PR to that effect. Double checking what storage does or doesn't get written to the ephemeral disk and whether the over-subscribed disk quota is effected by the root disk or not.

PR #243, opened 2024-09-16
Reduce storage requirements of a default deployment

lathiat:
Always deploy an 'ephemeral' storage disk for nova-compute, even if we are not 
using vault. While vault is pulled in by OVN in the majority of currently used
OpenStack versions, this will ensure we always have one of that size
regardless.

Remove the increased root-disk=80G size for nova-compute units, since we can 
use the space from the ephemeral disk instead.

Set a default root-disk=20GB constraint. Previously the prodstack6 environment 
would give us 20GB root disks as that is set in the default flavors, however
when booting from volume with root-disk-source=volume this is ignored and a
juju default of 30GB is used. Switch this to 20GB to go back to the previous
amount of storage.

With these changes the size of a jammy-yoga deployment with Ceph and 3x 
nova-compute nodes is reduced from 960GB to 600GB.

Closes: #229

### DIFF
--- a/common/generate_bundle_base
+++ b/common/generate_bundle_base
@@ -51,7 +51,7 @@ fi
 if [ -n "${MASTER_OPTS[MODEL_CONSTRAINTS]}" ]; then
     juju set-model-constraints ${MASTER_OPTS[MODEL_CONSTRAINTS]}
 else
-    juju set-model-constraints root-disk-source=volume
+    juju set-model-constraints root-disk-source=volume root-disk=20G
 fi
 
 if has_opt --list; then
--- a/openstack/module_defaults
+++ b/openstack/module_defaults
@@ -54,7 +54,7 @@ MOD_PARAMS[__PATH_MTU__]=
 
 # This is enough for creating one ubuntu vm or multiple cirros vms but some
 # scenarios may want to allow more per compute (e.g. octavia).
-MOD_PARAMS[__NOVA_COMPUTE_UNIT_CONSTRAINTS__]="mem=4G root-disk=80G"
+MOD_PARAMS[__NOVA_COMPUTE_UNIT_CONSTRAINTS__]="mem=4G cores=2"
 
 # Try to use current model (or newly requested one) as subdomain name
 model_subdomain=`get_juju_model`
--- a/openstack/openstack.yaml.template
+++ b/openstack/openstack.yaml.template
@@ -35,6 +35,8 @@ applications:
       migration-auth-type: ssh
       openstack-origin: *openstack_origin
       force-raw-images: false  # disable for stsstack since conversion kills the disks and is not needed
+    storage:
+      ephemeral-device: cinder,50G,1
   nova-cloud-controller:
     num_units: __NUM_NOVACC_UNITS__
     charm: __CHARM_STORE____CHARM_CS_NS____CHARM_CH_PREFIX__nova-cloud-controller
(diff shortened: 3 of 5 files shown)


openstack/module_defaults
```diff
@@ -54,7 +54,7 @@ MOD_PARAMS[__PATH_MTU__]=
 
 # This is enough for creating one ubuntu vm or multiple cirros vms but some
 # scenarios may want to allow more per compute (e.g. octavia).
-MOD_PARAMS[__NOVA_COMPUTE_UNIT_CONSTRAINTS__]="mem=4G root-disk=80G"
+MOD_PARAMS[__NOVA_COMPUTE_UNIT_CONSTRAINTS__]="mem=4G"
```
dosaboy, line 57: Im afraid this will break octavia which needs enough capacity to spawn an amphora vm and a member vm.

common/generate_bundle_base
```diff
@@ -51,7 +51,7 @@ fi
 if [ -n "${MASTER_OPTS[MODEL_CONSTRAINTS]}" ]; then
     juju set-model-constraints ${MASTER_OPTS[MODEL_CONSTRAINTS]}
 else
-    juju set-model-constraints root-disk-source=volume
+    juju set-model-constraints root-disk-source=volume root-disk=20G
```
dosaboy, line 54: i think this makes sense for the general case. Just need to make sure that vms that do actually need the extra capacity define it.

openstack/module_defaults
lathiat: The VMs are stored on the ephemeral storage disk, which has in practice almost always been attached at a size of 50GB anytime you used Vault which is most of the time as it's pulled in by OVN. This is due to `stoorage: ephemeral-device: cinder,50G,1` in overlays/openstack/vault-openstack-secrets.yaml.

The storage space from this mount point is what is reflected in the placement service/hypervisor stats when it's mounted, hence the 80GB root-disk size has mostly been irrelevant.

To handle non-OVN environments (either explicitly requested or for very old openstack versions where we didn't default to it), I also moved the `stroage: ephemeral-device: cinder,50G,1` from overlays/openstack/vault-openstack-secrets.yaml to openstack.yaml.template so that we would also always have this extra disk space in any deployment.

common/generate_bundle_base
lathiat: I agree, however it was mostly already working before we switched to root-disk-as-volume only a couple of weeks ago. So I think we're fine to ship this default and fix anything that comes up separately, unless you know of a specific case that needs fixing ahead of time.

The most obvious cases that come to mind are probably microk8s/kubernetes and maybe? mysql.

lathiat: I should actually quickly check, some instance types may have gotten more disk accidentally in the past due to flavor selection for ram or cores. I'll check on that.

openstack/module_defaults
dosaboy: my deployment seems to not be behaving that way. I have vault related to nova-compute:secret-storage but I still have an 80G root volume:

```
$ juju ssh nova-compute/0 df -h| egrep '/dev/vda1 '
/dev/vda1        78G  3.3G   75G   5% /
```

dosaboy: I assume that juju is picking the greater of the two constraints

dosaboy: I see your point now, nova is using the space available in /varlib/nova/instances not /

common/generate_bundle_base
lathiat: I tested this out, for a ceph and octavia deployment the only machine that got more cores/ram/storage by accident of prodstack6 flavors was nova-compute, so I pushed a new revision which also specifies cores=2 to ensure we still get the same.

lathiat:
Pushed a new revision which also specifies cores=2 for nova constraints as we used to get that as a result of the only available flavours for ram=4G/ram=8G. I believe this is ready to merge now.

dosaboy (approved):
I think this all makes sense now thanks for putting in the time

PR #245, opened 2024-09-16
charmed functest runner follow up improvements

rodrigogansobarbieri:


openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -42,7 +42,7 @@ OPTIONS:
     --manual-functests
         Runs functest commands separately (deploy,configure,test) instead of
         the entire suite.
-    --remote-build USER@HOST,GIT_PATH
```
dosaboy, line 45: Can we stick with the UPPERCASE format? its a widely used standard for cli tools

dosaboy (approved):
LGTM!

PR #246, opened 2024-09-17
Fix func target voting test

dosaboy:
We need to support job defs that have not voting key which implies voting=True.

Also adds __pycache__ to .gitignore


rodrigogansobarbieri (approved):
LGTM

PR #252, opened 2024-09-25
Ensure func tests log both stdout and stderr

dosaboy:
Juju 2 prints output to stderr as opposed to stdout in Juju 3 so need to pipe both to logs.


pponnuvel (approved):
LGTM

PR #253, opened 2024-09-27
Support func test zaza overlays

dosaboy:
Some charms define overlays in tests.yaml that map to a job in osci.yaml.

### DIFF
--- a/openstack/tools/func_test_tools/identify_charm_func_tests.py
+++ b/openstack/tools/func_test_tools/identify_charm_func_tests.py
@@ -22,14 +22,22 @@ def extract_targets(bundle_list):
     extracted = []
     for item in bundle_list or []:
         if isinstance(item, dict):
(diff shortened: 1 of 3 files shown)


openstack/tools/func_test_tools/identify_charm_func_tests.py
```diff
@@ -22,14 +22,21 @@ def extract_targets(bundle_list):
     extracted = []
     for item in bundle_list or []:
         if isinstance(item, dict):
-            extracted.append(list(item.values())[0])
+            values = list(item.values())
+            if isinstance(values, list):
```
pponnuvel, line 26: Will this ever be false?

dosaboy: yeah this is clearly wrong i need to fix it

PR #258, opened 2024-10-02
Cleanup old zaza models

dnegreira:
It can happen that when running the functests that the zaza model stays behind, either by cancelling the script or if the script has an error.

This patch adds a cleanup check to verify if there are old zaza-* models before proceeding to run the functests.

### DIFF
--- a/openstack/tools/charmed_openstack_functest_runner.sh
+++ b/openstack/tools/charmed_openstack_functest_runner.sh
@@ -110,6 +110,15 @@ done
 # Install dependencies
 which yq &>/dev/null || sudo snap install yq
 
+# Cleanup zaza-* models before proceeding
+OLD_ZAZA_MODELS=$(juju list-models| grep -E "^zaza-\S+"|tr -d '*')
+if [ -n "${OLD_ZAZA_MODELS}" ]; then
+    echo "Old zaza-* models active, cleaning up."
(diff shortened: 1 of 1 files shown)


openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -110,6 +110,15 @@ done
 # Install dependencies
 which yq &>/dev/null || sudo snap install yq
 
+# Cleanup zaza-* models before proceeding
+OLD_ZAZA_MODELS=$(juju list-models| grep -E "^zaza-\S+"|tr -d '*')
+if [ -n "${OLD_ZAZA_MODELS}" ]; then
```
dosaboy, line 115: there is a helper function to do this now - https://github.com/canonical/stsstack-bundles/blob/main/openstack/tools/func_test_tools/common.sh#L1

it is already being run before and after a test run, do we really need to do it again here?

dosaboy: see https://github.com/canonical/stsstack-bundles/commit/f0a30975f3bd9d9875bd984a1f8726fdf0879cbd

dnegreira: well, this was not there when I wrote it yesterday, so I guess this PR can be dropped

PR #262, opened 2024-10-03
Func tests branch detection fallback

dosaboy:
.gitreview will not have a defaultbranch config
on master branch.


wilkmar (approved):
LGTM

PR #263, opened 2024-10-03
Use local project functioan test template

dosaboy:
Need to make sure we use the local functional test template when looking up voting info in zosci-config. Falls back to default charm-functional-jobs if no local one found.


dnegreira (approved):

openstack/tools/func_test_tools/common.py
```diff
@@ -56,13 +56,11 @@ def get_branch_jobs(self, branch, project_templates):
 class OSCIConfig():
     """ Extract information from osci.yaml """
     def __init__(self):
-        if not os.path.exists('osci.yaml'):
-            self._osci_config = {}
-        else:
-            with open('osci.yaml', encoding='utf-8') as fd:
-                self._osci_config = yaml.safe_load(fd)
+        path = os.path.join(os.environ.get('CHARM_ROOT_PATH', ''), 'osci.yaml')
+        with open(path, encoding='utf-8') as fd:
```
dnegreira, line 60: When opening a file we should have in an exception catcher so that we can handle it properly

Something like:
```
try:
    with open(path, encoding='utf-8') as fd:
        self._osci_config = yaml.safe_load(fd)
except Exception as e:
    print(f"Error opening {file}: {e}")

dosaboy: no i actually want it to fail hard since if it cant find this information it cant do its job properly (which is what happened before)

dnegreira (approved):

PR #268, opened 2024-10-07
revert to cirros-0.4.0

### LINKED ISSUE #267: revert to cirros-0.4.0
I tested all images from cirros 0.6.3 to 0.4.0 and 0.4.0 is the latest one that works. By works I mean, it boots and print console log messages.

rodrigogansobarbieri:
closes #267


dosaboy (approved):
Would be good to get a better understanding of why those other versions dont work but since this one does im happy to go with it for now.

Issue #269, opened 2024-10-08
Need to stop OpenStack allocating neutron ports from the "VIP range"


lathiat:
This is required by both stsstack-bundles and the charmed test runner

A couple of possible solutions

1. Allocate intentionally un-used undercloud neutron ports with the VIPs we plan to use - watch out for accidentally cleaning them up
2. Modify the neutron subnet allocation_range to remove these IPs. I tried to do this in terraform, the problem is the subnet range is dynamically allocated by OpenStack during the subnet creation, so terraform cannot calculate the allocation_range because its a circular dependency. We can however still manually modify this ourselves, and terraform probably won't roll it back, because it's not tracking the value.

PR #271, opened 2024-10-08
Pre-allocate func test vip ports

dosaboy:
Resolves: #269

### DIFF
--- a/openstack/tools/charmed_openstack_functest_runner.sh
+++ b/openstack/tools/charmed_openstack_functest_runner.sh
@@ -138,9 +138,19 @@ OAM_MAX=$(ipcalc $CIDR_OAM| awk '$1=="HostMax:" {print $2}')
 OAM_MIN=$(ipcalc $CIDR_OAM| awk '$1=="HostMin:" {print $2}')
 OAM_MIN_ABC=${OAM_MIN%.*}
 OAM_MAX_D=${OAM_MAX##*.}
-# Picking last two addresses and hoping they dont get used by Neutron.
-export {OS,TEST}_VIP00=${OAM_MIN_ABC}.$(($OAM_MAX_D - 1))
-export {OS,TEST}_VIP01=${OAM_MIN_ABC}.$(($OAM_MAX_D - 2))
+
+NUM_VIPS=2
+for ((i=$NUM_VIPS;i;i-=1)); do
+    vip_port_name=zaza-vip-port-$((i-1))
+    vip_addr=$(openstack port show -c fixed_ips $vip_port_name -f yaml| yq .fixed_ips[0].ip_address)
+    if [[ $vip_addr = null ]]; then
+        vip_addr=${OAM_MIN_ABC}.$(($OAM_MAX_D - $i))
+        # Pre-allocate ports with addresses used for VIPs so that they don't
+        # collide with the deployment itself.
+        allocate_port $vip_addr net_${OS_USERNAME}-psd $vip_port_name
+    fi
+    export {OS,TEST}_VIP0$((i-1))=$vip_addr
+done
 
 # More information on config https://github.com/openstack-charmers/zaza/blob/master/doc/source/runningcharmtests.rst
 export {,TEST_}NET_ID=$(openstack network show net_${OS_USERNAME}-psd-extra -f value -c id)
--- a/openstack/tools/func_test_tools/common.sh
+++ b/openstack/tools/func_test_tools/common.sh
@@ -44,3 +44,12 @@ apply_func_test_pr ()
         -f './src/test-requirements*.txt' \
         "$msg"
 }
+
+allocate_port ()
+{
+    local addr=$1
+    local net_name=$2
+    local port_name=$3
+    openstack port create --network $net_name --fixed-ip ip-address=$addr \
+        $port_name
+}


rodrigogansobarbieri:
Who de-allocates the ports? Also, doesn't commit [1] prevent the issue altogether?

[1] https://github.com/canonical/stsstack-bundles/commit/50e87ed37e48cbd168827a9096a7d58c618f2684

dosaboy:
> Who de-allocates the ports? Also, doesn't commit [1] prevent the issue altogether?
> 
> [1] [50e87ed](https://github.com/canonical/stsstack-bundles/commit/50e87ed37e48cbd168827a9096a7d58c618f2684)

do we want to de-allocate the ports? i mean im fine to do so but just not sure its necessary since they will always use the same address. now that you mention it though this will potentially cause issues with the main vip code used for deployments so i guess de-allocation makes sense.

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -138,9 +138,19 @@ OAM_MAX=$(ipcalc $CIDR_OAM| awk '$1=="HostMax:" {print $2}')
...
-# Picking last two addresses and hoping they dont get used by Neutron.
-export {OS,TEST}_VIP00=${OAM_MIN_ABC}.$(($OAM_MAX_D - 1))
-export {OS,TEST}_VIP01=${OAM_MIN_ABC}.$(($OAM_MAX_D - 2))
+
+NUM_VIPS=2
+for ((i=$NUM_VIPS;i;i-=1)); do
+    vip_port_name=zaza-vip-port-$((i-1))
+    vip_addr=$(openstack port show -c fixed_ips $vip_port_name -f yaml| yq .fixed_ips[0].ip_address)
+    if [[ $vip_addr = null ]]; then
+        vip_addr=${OAM_MIN_ABC}.$(($OAM_MAX_D - $i))
```
brianphaley, line 147: Does the IP address matter? It shouldn't, seems this could cause an issue if the IP is already in use.

dosaboy: thats a good point, we could just use address now couldnt we since we are pre-allocating them. nice one brian.

dosaboy:
I have reviews the tools_allocate_vips.sh script. Currently it is not called from anywhere so is not integrated. It is also use specific IP addresses which as @brianphaley points out is no longer necessary but in fixing that in the rest of stsstack-bundles is going to be a not-so-insignificant patch so what I have done is used a port naming convention compatible with that script and will follow up with a patch to fix the wider issue which when done will mean we can remove this implementation (which is now much simpler than before btw).

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -133,14 +133,20 @@ FIP_MIN_ABC=${FIP_MIN%.*}
...
-OAM_MIN=$(ipcalc $CIDR_OAM| awk '$1=="HostMin:" {print $2}')
-OAM_MIN_ABC=${OAM_MIN%.*}
-OAM_MAX_D=${OAM_MAX##*.}
-# Picking last two addresses and hoping they dont get used by Neutron.
-export {OS,TEST}_VIP00=${OAM_MIN_ABC}.$(($OAM_MAX_D - 1))
-export {OS,TEST}_VIP01=${OAM_MIN_ABC}.$(($OAM_MAX_D - 2))
+# Setup VIPs
+NUM_VIPS=2
+for ((i=$NUM_VIPS;i;i-=1)); do
+    # We use the same naming convention as ../tools/allocate_vips.sh to avoid conflicts and re-use
```
rodrigogansobarbieri, line 139: The logic here is fine and I'm fine with it, but previously I had suggested using what already exists instead. This logic is creating 2 VIPs, the tools/allocate_vips.sh script creates 20. I am not sure if the tools/allocate_vips.sh script will fail if 2 are already allocated. Sounds easier to me if you just do a openstack port list | grep ps6-vip and if there is any result then you skip, otherwise you run tools/allocate_vips.sh to allocate 20.

openstack/tools/openstack_regression_tests_runner.sh
```diff
@@ -90,14 +90,20 @@ FIP_MIN_ABC=${FIP_MIN%.*}
 FIP_MIN_D=${FIP_MIN##*.}
 FIP_MIN=${FIP_MIN_ABC}.$(($FIP_MIN_D + 64))
 
-CIDR_OAM=$(openstack subnet show subnet_${OS_USERNAME}-psd -c cidr -f value)
```
rodrigogansobarbieri, line 93: I didn't realize the code was duplicated. Could be a good future idea to move this code to a function in a single file that is used by both scripts if there is nothing preventing that. Something for the future improvements, not now.

rodrigogansobarbieri (commented):
comments inline, not -1 nor +1

openstack/tools/charmed_openstack_functest_runner.sh
rodrigogansobarbieri: I forgot to ask in the above comment: had you considered that approach I suggested? I'm thinking maybe you did and then you implemented as you are proposing for a specific reason.

dosaboy: actually my point was more that the allocatte_vips script is creating with specific addresses which as brian pointed out is not necessary and prone to conflicts at create time but fixing it will be too much work for this patchset so I will followup

openstack/tools/openstack_regression_tests_runner.sh
dosaboy: +1

openstack/tools/charmed_openstack_functest_runner.sh
rodrigogansobarbieri: good point, ideally everyone should run allocate_vips before they have any running resources, but if they don't then it would cause problems.

openstack/tools/func_test_tools/common.sh
```diff
@@ -44,3 +44,30 @@ apply_func_test_pr ()
...
+    # We use the same naming convention as ../tools/allocate_vips.sh to avoid conflicts and re-use
+    # those vips.
+    vip_port_name=ps6-vip-ip$((vip_id-1))
+    vip_addr=$(openstack port show -c fixed_ips $vip_port_name -f yaml| yq .fixed_ips[0].ip_address)
+    if [[ $vip_addr = null ]]; then
+        # Pre-allocate ports with addresses used for VIPs so that they don't
+        # collide with the deployment itself.
+        allocate_port net_${OS_USERNAME}-psd $vip_port_name
+    fi
+    echo $vip_addr
```
brianphaley, line 72: vip_addr will not be valid here if we created a port - I think you can get the value by using -c -f with 'port create' but haven't tried it.

brianphaley (changes requested):

openstack/tools/func_test_tools/common.sh
dosaboy: thank you, will fix

openstack/tools/func_test_tools/common.sh
```diff
@@ -44,3 +44,31 @@ apply_func_test_pr ()
...
+create_zaza_vip ()
+{
+    # Allocates a vip ensuring to use existing ones if they exist.
+    #
+    # Returns the address of the vip.
+    #
+    local vip_id=$1
+    # We use the same naming convention as ../tools/allocate_vips.sh to avoid conflicts and re-use
+    # those vips.
+    vip_port_name=ps6-vip-ip$((vip_id-1))
```
brianphaley, line 66: Is this vip_id-1 correct? I think the caller passes (1, 0)

> for ((i=2;i;i-=1)); do echo $((i-1)); done
1
0

brianphaley (changes requested):

openstack/tools/func_test_tools/common.sh
dosaboy: actually the caller passes $i but would make more sense if it passed $i-1 so ill fix it

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -133,14 +133,10 @@ FIP_MIN_ABC=${FIP_MIN%.*}
...
-OAM_MIN=$(ipcalc $CIDR_OAM| awk '$1=="HostMin:" {print $2}')
-OAM_MIN_ABC=${OAM_MIN%.*}
-OAM_MAX_D=${OAM_MAX##*.}
-# Picking last two addresses and hoping they dont get used by Neutron.
-export {OS,TEST}_VIP00=${OAM_MIN_ABC}.$(($OAM_MAX_D - 1))
-export {OS,TEST}_VIP01=${OAM_MIN_ABC}.$(($OAM_MAX_D - 2))
+# Setup vips needed by zaza tests.
+for ((i=2;i;i-=1)); do
+    export {OS,TEST}_VIP0$((i-1))=$(create_zaza_vip (($i-1)))
+done
```
rodrigogansobarbieri, line 139: /home/ubuntu/stsstack-bundles/openstack/tools/charmed_openstack_functest_runner.sh: command substitution: line 139: syntax error near unexpected token `('

rodrigogansobarbieri (changes requested):

openstack/tools/charmed_openstack_functest_runner.sh
dosaboy: omg

dosaboy: why didnt bashate find that

dosaboy: anyway, fixed.

openstack/tools/func_test_tools/common.sh
```diff
@@ -44,3 +44,31 @@ apply_func_test_pr ()
...
+    # Allocates a vip ensuring to use existing ones if they exist.
+    #
+    # Returns the address of the vip.
+    #
+    local vip_id=$1
+    # We use the same naming convention as ../tools/allocate_vips.sh to avoid conflicts and re-use
+    # those vips.
+    vip_port_name=ps6-vip-ip$vip_id
+    vip_addr=$(openstack port show -c fixed_ips $vip_port_name -f yaml| yq .fixed_ips[0].ip_address)
+    if [[ $vip_addr = null ]]; then
```
rodrigogansobarbieri, line 68: What if the ps6-vip-ip01 and ps6-vip-ip02 exist but are in use? I would need to review the generate_bundles script to make sure whether those 2 first IPs are likely to be chosen by it or not. It may look a bit complicated to coordinate sharing those VIPs with the ones pre-allocated by the ../tools/allocate_vips.sh script so it is best if we just make sure to adjust the range in the generate_bundles script to use the 18 last IPs, and keep ip01 and ip02 off its reach. That's all.

rodrigogansobarbieri (changes requested):

openstack/tools/func_test_tools/common.sh
dosaboy: this assumes they are not in use - we generally dont have quota for multiple deployments, its pretty unlikely that someone will have more than one HA deployment on the go at once. can fix later if needs be

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -133,14 +133,10 @@ FIP_MIN_ABC=${FIP_MIN%.*}
...
-OAM_MAX=$(ipcalc $CIDR_OAM| awk '$1=="HostMax:" {print $2}')
-OAM_MIN=$(ipcalc $CIDR_OAM| awk '$1=="HostMin:" {print $2}')
-OAM_MIN_ABC=${OAM_MIN%.*}
-OAM_MAX_D=${OAM_MAX##*.}
-# Picking last two addresses and hoping they dont get used by Neutron.
-export {OS,TEST}_VIP00=${OAM_MIN_ABC}.$(($OAM_MAX_D - 1))
-export {OS,TEST}_VIP01=${OAM_MIN_ABC}.$(($OAM_MAX_D - 2))
+# Setup vips needed by zaza tests.
+for ((i=2;i;i-=1)); do
+    export {OS,TEST}_VIP0$((i-1))=$(create_zaza_vip $(($i-1)))
```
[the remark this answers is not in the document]
dosaboy: will fix

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -133,14 +133,10 @@ FIP_MIN_ABC=${FIP_MIN%.*}
...
-OAM_MAX=$(ipcalc $CIDR_OAM| awk '$1=="HostMax:" {print $2}')
-OAM_MIN=$(ipcalc $CIDR_OAM| awk '$1=="HostMin:" {print $2}')
-OAM_MIN_ABC=${OAM_MIN%.*}
-OAM_MAX_D=${OAM_MAX##*.}
-# Picking last two addresses and hoping they dont get used by Neutron.
-export {OS,TEST}_VIP00=${OAM_MIN_ABC}.$(($OAM_MAX_D - 1))
-export {OS,TEST}_VIP01=${OAM_MIN_ABC}.$(($OAM_MAX_D - 2))
+# Setup vips needed by zaza tests.
+for ((i=2;i;i-=1)); do
+    export {OS,TEST}_VIP0$((i-1))=$(create_zaza_vip 0$(($i-1)))
```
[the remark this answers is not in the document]
dosaboy: this gives OS_VIP00=VIP1, OS_VIP00=VIP2 etc which is what we want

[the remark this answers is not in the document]
dosaboy: the addresses are dynamic and based on whatever neutron allocates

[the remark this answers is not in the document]
dosaboy: at least if we use ```OS_VIP00=ps6-vip-ip00``` then the numbers either side match which makes it easier to match the variable to the port name? I dont see the value in making them different?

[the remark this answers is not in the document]
dosaboy: see reply in the dup of your comment above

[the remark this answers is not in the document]
dosaboy: ok i have changed it now so that it is more inline with tools/preallocate_vips.sh

rodrigogansobarbieri (approved):

PR #272, opened 2024-10-11
Adding nova-compute machine constraints root-disk=80G

xtrusia:
Current code that adding root-disk=80G to nova-compute application definition is not properly working with charm-manila-ganesha and with 'services' bundle entity

I assume that 'services' is old entity which doesn't support constraints while 'applications' does

I upload this patch to add root-disk to machine constraints instead for compatability.

### DIFF
--- a/openstack/tools/charmed_openstack_functest_runner.sh
+++ b/openstack/tools/charmed_openstack_functest_runner.sh
@@ -218,11 +218,12 @@ if $MODIFY_BUNDLE_CONSTRAINTS; then
     for f in tests/bundles/*.yaml; do
         # Dont do this if the test does not have nova-compute
         if $(grep -q "nova-compute:" $f); then
-            if [[ $(yq '.applications' $f) = null ]]; then
-                yq -i '.services.nova-compute.constraints="root-disk=80G mem=8G"' $f
-            else
-                yq -i '.applications.nova-compute.constraints="root-disk=80G mem=8G"' $f
-            fi
+            # Taking nova-compute machines and add root-disk=80G to machine constraints
+            MACHINES=($(yq eval '(.services.nova-compute.to[] // .applications.nova-compute.to[])' "$f"))
+
+            for MACHINE in "${MACHINES[@]}"; do
+                yq eval -i ".machines[\"${MACHINE}\"].constraints = \"root-disk=80G mem=8G\"" "$f"
+            done
         fi
     done
     )


openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -218,11 +218,12 @@ if $MODIFY_BUNDLE_CONSTRAINTS; then
     for f in tests/bundles/*.yaml; do
         # Dont do this if the test does not have nova-compute
         if $(grep -q "nova-compute:" $f); then
-            if [[ $(yq '.applications' $f) = null ]]; then
-                yq -i '.services.nova-compute.constraints="root-disk=80G mem=8G"' $f
-            else
-                yq -i '.applications.nova-compute.constraints="root-disk=80G mem=8G"' $f
-            fi
+            # Taking nova-compute machines and add root-disk=80G to machine constraints
+            MACHINES=($(yq eval '(.services.nova-compute.to[] // .applications.nova-compute.to[])' "$f"))
```
pponnuvel, line 222: Hi Seyeong,

I am not sure if you intend to have a `.to` there as that's not a field in typical bundles.

Also I don't understand how this is intended to work. `constraints` we're looking to change here is a per application field, right? So what's the idea behind getting a machine list and changing?

In a bundle I've, `nova-compute` section looks like:
```
  nova-compute:
    charm: ch:nova-compute
    channel: yoga/edge
    series: focal
    num_units: 1
    constraints: mem=4G cores=2
    options:
      debug: *debug
      verbose: *verbose
      enable-live-migration: True
      enable-resize: True
      migration-auth-type: ssh
      openstack-origin: *openstack_origin
      force-raw-images: false  # disable for stsstack since conversion kills the disks and is not needed
    storage:
      ephemeral-device: cinder,50G,1
```

This produces nothing:
```
$ yq eval '(.services.nova-compute.to[] // .applications.nova-compute.to[])' openstack.yaml
```

And
```
$ yq eval '(.services.nova-compute.[] // .applications.nova-compute.[])' openstack.yaml 
```
(assuming `.to` is a mistake) produces:
```
ch:nova-compute
yoga/edge
focal
1
mem=4G cores=2
debug: *debug
verbose: *verbose
enable-live-migration: True
enable-resize: True
migration-auth-type: ssh
openstack-origin: *openstack_origin
force-raw-images: false # disable for stsstack since conversion kills the disks and is not needed

ephemeral-device: cinder,50G,1
```

So I don't understand how this is supposed to work. Perhaps paste a bundle for which the current code fails.

xtrusia: oh right. I referred to this bundle
https://opendev.org/openstack/charm-manila-ganesha/src/branch/stable/2023.1/src/tests/bundles/jammy-antelope.yaml

recently I'm working on this charm and current charm test code is not working( I have to put root-disk=80G to machine constraints. and it works. )

but if normally there is no .to, then... this is not the common case as you said.

maybe changing 'services' above bundle's to 'applications' could be the better solution.

PR #273, opened 2024-10-31
add k8s bootstrap script

slapcat:
This script is a building block for moving from a machine model deployed with the kubernetes/microk8s bundle to a k8s cloud with native juju controller so that k8s bundles can be deployed. It simplifies the mechanics of the only existing k8s bundle (OSM) and lays the groundwork for the Canonical IAM bundle I am working on.

The script also supports Canonical K8s even though we don't have a bundle for that yet.


slapcat:
Hi @dosaboy - this is for the IAM bundle we discussed before. Can you please review?

nicolasbock (approved):
LGTM

slapcat:
@nicolasbock Looks like this was approved but still open. Can you merge it?

PR #275, opened 2024-11-15
fix setting OS_AUTH_PROTOCOL to https

wilkmar:
The OS_AUTH_PROTOCOL should be set to 'https' only when a relation of type 'certificates' exists between keystone and vault. Before it was set when the the vault charm was found in the deployment.

### DIFF
--- a/openstack/novarc
+++ b/openstack/novarc
@@ -40,8 +40,8 @@ END
                     return;;
             esac
         fi
+        export OS_AUTH_PROTOCOL=https
     fi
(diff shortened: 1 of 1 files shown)


wilkmar:
@dosaboy can you have a look at this one pls?

dosaboy (approved):
I'm ok with this because it will still work and also support the scenario where vault is deployed but not used for api SSL but that is not actually a possible configuration with stsstack-bundles - we always deploy with api ssl. So this is to support your customised bundles where vault is only used for ovn certs.

PR #279, opened 2024-12-03
Don't set root-disk-source when using MAAS

dosaboy:


common/generate_bundle_base
```diff
@@ -50,7 +50,7 @@ fi
 
 if [ -n "${MASTER_OPTS[MODEL_CONSTRAINTS]}" ]; then
     juju set-model-constraints ${MASTER_OPTS[MODEL_CONSTRAINTS]}
-else
+elif ! ${MASTER_OPTS[HYPERCONVERGED_DEPLOYMENT]}; then
```
nicolasbock, line 53: We are not setting any constraints in the `else` case? I am not saying that we
should, but that's a change in behavior.

brianphaley, line 53: nit: not sure if the check should use "" in case the option is empty like above?

elif [ -n "${MASTER_OPTS[HYPERCONVERGED_DEPLOYMENT]}" ]; then

dosaboy: there are no constraints to set are there?

dosaboy: HYPERCONVERGED_DEPLOYMENT defaults to false (bool) and is only ever set to true otherwise

nicolasbock: I don't think so. I just wanted to make sure I am not missing anything.

nicolasbock: LGTM

nicolasbock (approved):

PR #283, opened 2024-12-06
Fix upload octavia script

nicolasbock:

### DIFF
--- a/openstack/tools/upload_octavia_amphora_image.sh
+++ b/openstack/tools/upload_octavia_amphora_image.sh
@@ -1,2 +1,46 @@
-#!/bin/bash -eu
-juju run octavia-diskimage-retrofit/0 retrofit-image
+#!/bin/bash
+
+set -e -u
+
+declare amphora_series=jammy
+
+while (( $# > 0 )); do
+    case $1 in
+        -h|--help)
+            cat <<EOF
+Usage:
+
+-h | --help         This help
+--series SERIES     The series (also the image name) to use as the amphora base
+EOF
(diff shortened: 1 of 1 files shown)


openstack/tools/upload_octavia_amphora_image.sh
```diff
@@ -1,2 +1,23 @@
...
+#!/bin/bash
+
+set -e -u
+
+basedir=$(realpath $(dirname $0))
+source ${basedir}/../novarc
+
+set -x
+
+ID=$(openstack image show --format value --column id jammy)
```
brianphaley, line 10: My only enhancement here would be to have the script accept an optional argument - focal/jammy/noble/etc, but default to jammy if none is given.

nicolasbock: I added a command line option for the series.

brianphaley (approved):

nicolasbock:
The command is failing with:

```console
$ ./tools/upload_octavia_amphora_image.sh 
Running operation 5 with 1 task
  - task 6 on unit-octavia-diskimage-retrofit-0

Waiting for task 6...
Action id 6 failed: unable to find suitable source image
```

I updated the commit message.

nicolasbock (commented):
The `retrofit` action also times out without the `--wati= 0s` command line option and it fails to add a tag to the image it produces.

nicolasbock:
The retrofit action is actually working. The issue was an overly aggressive approach in `configure`, see  #291 .

Issue #295, opened 2025-01-09
openstack releases on noble should use latest/edge channel instead of 2024.1


rodrigogansobarbieri:
there is no build of charms on noble on any channel except for latest/edge

PR #296, opened 2025-01-09
Add oracular/dalmatian support

rodrigogansobarbieri:
Also fixes charm channels on noble and oracular/dalmatian to use latest/edge.

Closes: #294
Closes: #295


common/ch_channel_map/any_series
```diff
@@ -3,9 +3,15 @@ declare -A RELNAME_MAP=(
     [antelope]=2023.1
     [bobcat]=2023.2
     [caracal]=2024.1
+    [dalmatian]=latest
```
brianphaley, line 6: dalmatian is 2024.2

common/openstack_release_info
```diff
@@ -24,12 +24,45 @@ declare -A lts=( [trusty]=icehouse
                  [jammy]=yoga
                  [noble]=caracal
 )
+
+# These hardcoded sorted lists replace the previous "sorted" lists
+# because we cannot sort when the alphabet wraps around.
+declare -a lts_releases_sorted=( caracal
+                                 yoga
```
nicolasbock, line 31: The indentation is a bit off.

common/ch_channel_map/any_series
dosaboy: at present the only way to deploy noble (to get D and E uca) is to use latest/edge charms

common/ch_channel_map/oracular
```diff
@@ -0,0 +1,18 @@
...
+
+# "Charm Delivery" specifies vault 1.8 for jammy, however field have actually been using 1.7.
+# Stick with 1.7 to more closely replicate customer deployments.
+CHARM_CHANNEL[vault]=latest/edge
+
+CHARM_CHANNEL[rabbitmq-server]=latest/edge
+CHARM_CHANNEL[mysql-router]=latest/edge
+CHARM_CHANNEL[mysql-innodb-cluster]=latest/edge
+CHARM_CHANNEL[hacluster]=2.4/edge
+CHARM_CHANNEL[ovn-central]=23.09/edge # FIXME: update once caracal released
```
dosaboy, line 16: need to fix these since the correct version is 24.03. I will submit a separate patch.

common/ch_channel_map/oracular
```diff
@@ -0,0 +1,18 @@
+# NOTE(ganso): This was copy-pasted from noble
```
dosaboy, line 1: maybe use a symbolic link rather than copying? then we dont have to update multiple files.

common/ch_channel_map/oracular
dosaboy: https://github.com/canonical/stsstack-bundles/pull/299

common/openstack_release_info
nicolasbock: Yoga is still indented but starting from ussuri the indentation is too far left.

nicolasbock: Maybe an issue with tab vs. space?

common/ch_channel_map/oracular
dosaboy: right but the point is not to have duplicate files so use symlink until they need to diverge

common/openstack_release_info
nicolasbock: Thanks, that looks better.

nicolasbock (approved):

dosaboy (approved):
lgtm

PR #297, opened 2025-01-09
Also remove volume that have different names

nicolasbock:
The 'in-use' volumes, attached to non-existent servers, do not
necessarily have `juju` in their name. This change removes this
restriction.

### DIFF
--- a/tools/cleanup-cinder-attachments.sh
+++ b/tools/cleanup-cinder-attachments.sh
@@ -1,5 +1,6 @@
 #!/bin/bash -u
-for vol in $(openstack volume list| grep juju| grep in-use| awk '{print $2}'); do
+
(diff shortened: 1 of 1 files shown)


tools/cleanup-cinder-attachments.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash -u
-for vol in $(openstack volume list| grep juju| grep in-use| awk '{print $2}'); do
+
```
dosaboy, line 2: If we are going to ignore name we should do ```openstack volume list --status in-use```

nicolasbock: Yes, that is simpler.

dosaboy (approved):

PR #301, opened 2025-01-17
Add 1GB flavor for ubuntu images

rodrigogansobarbieri:
1GB is ideal for testing ubuntu images, 2GB was too much considering our compute VMs have only 4GB of RAM. This is essential when testing octavia because you need enough RAM for the user VM, the amphora, and the compute host to avoid crashes.

The xsmall flavor proposed is the same as the amphora size.


dosaboy (approved):
lgtm

Issue #304, opened 2025-04-03
OpenStack ./configure is not working


sombrafam:
When I try to go figure a network in my ovs deployment I get this:

```
+ '[' -n vxlan ']'
+ '[' -n vxlan ']'
++ juju config neutron-api enable-dvr
+ dvr_enabled=true
+ '[' true = true ']'
+ false
+ ./bin/add-data-ports.sh '' 'net_"stg-reproducer-sombrafam-psd-extra"'
Managing ports for neutron-openvswitch units
Checking neutron-openvswitch unit instances: a9d0b754-2e7d-4d92-a22e-39711ca46489 7fc9983f-8068-4b48-a1fe-ceca97062d52 af3a7593-5396-4860-aa3e-792089f83813
No Network found for net_"stg-reproducer-sombrafam-psd-extra"
No Network found for net_"stg-reproducer-sombrafam-psd-extra"
No Network found for net_"stg-reproducer-sombrafam-psd-extra"
No Network found for net_"stg-reproducer-sombrafam-psd-extra"
Creating 3 new ports
No Network found for net_"stg-reproducer-sombrafam-psd-extra"
```

This happens because of the way the network name is being fetched and the quotes are not removed.

sombrafam:
How to reproduce the issue:

```./generate-bundle.sh --series jammy --release yoga --name jammy-yoga-l3 --openstack-dashboard --dvr-snat-l3ha  --ml2dns --ml2-ovs
./configure prodstack6
```

PR #305, opened 2025-04-03
Fix openstack ./configure

sombrafam:
The network names are not correctly being  expanded, and contains "", causing the configure script to fail.

Resolves issue #304


openstack/profiles/prodstack5
```diff
@@ -24,7 +24,7 @@ export GATEWAY=`openstack subnet show $EXT_SUBNET -c gateway_ip -f value`
 [[ -z "$FIP_RANGE" ]] && export FIP_RANGE="${FIP_RANGE_FIRSTIP}:${FIP_RANGE_LASTIP}"
 [[ -z "$CIDR_PRIV" ]] && export CIDR_PRIV="192.168.21.0/24"
 
-export UNDERCLOUD_OAM_NET="$(sed -E --quiet "s/.+OS_PROJECT_NAME=(.+)_project/net_\1-psd/p" ~/novarc)"
-export UNDERCLOUD_EXT_NET="$(sed -E --quiet "s/.+OS_PROJECT_NAME=(.+)_project/net_\1-psd-extra/p" ~/novarc)"
+export UNDERCLOUD_OAM_NET="$(source ~/novarc && echo net_${OS_USERNAME}-psd)"
+export UNDERCLOUD_EXT_NET="$(source ~/novarc && echo net_${OS_USERNAME}-psd-extra)"
```
brianphaley, line 28: I think the way you've done things is fine, but I'm curious of the change to OS_USERNAME is correct? Typically it is the same as OS_PROJECT_NAME, but you can have a single user with multiple projects, in which case you'd want to use the project name. I don't think we every do this in this repo though.

sombrafam: There might be have some changes on IS scrips  from the first time this script was made. Maybe it was right in the beginning, but now if you look at the network names, using the project name won't work anymore:


```
ubuntu@stg-reproducer-sombrafam-project-bastion:~$ export | grep OS_ | grep sombrafam
declare -x OS_PROJECT_NAME="stg-reproducer-sombrafam_project"
declare -x OS_USERNAME="stg-reproducer-sombrafam"
ubuntu@stg-reproducer-sombrafam-project-bastion:~$ openstack network list| grep sombrafam
| 33bf5813-5909-4468-ac30-162caedc62a2 | net_stg-reproducer-sombrafam-psd       | 0fa54984-e1a9-42b2-ac24-625b73d598c7                                       |
| 78bd0b4e-3094-469a-b473-b2ffd322f902 | net_stg-reproducer-sombrafam-pse       | bb4e4a3d-56cb-4cbd-ac37-aeba808f0321                                       |
| e7c117ca-e00e-473b-9968-fe9c20ae4ef7 | net_stg-reproducer-sombrafam-psd-extra | 909e48a4-c8c6-4d63-a4ee-0a24a49781cf                                       |
ubuntu@stg-reproducer-sombrafam-project-bastion
```

brianphaley: Oh, yuck, guess that's the way it has to be then. Another thing I just noticed is that  ~/novarc is already sourced, is there a need to do it on every line here?

openstack/profiles/prodstack6
```diff
@@ -24,7 +24,7 @@ export GATEWAY=`openstack subnet show $EXT_SUBNET -c gateway_ip -f value`
 [[ -z "$FIP_RANGE" ]] && export FIP_RANGE="${FIP_RANGE_FIRSTIP}:${FIP_RANGE_LASTIP}"
 [[ -z "$CIDR_PRIV" ]] && export CIDR_PRIV="192.168.21.0/24"
 
-export UNDERCLOUD_OAM_NET="$(sed -E --quiet "s/.+OS_PROJECT_NAME=(.+)_project/net_\1-psd/p" ~/novarc)"
-export UNDERCLOUD_EXT_NET="$(sed -E --quiet "s/.+OS_PROJECT_NAME=(.+)_project/net_\1-psd-extra/p" ~/novarc)"
+export UNDERCLOUD_OAM_NET="$(echo net_${OS_USERNAME}-psd)"
+export UNDERCLOUD_EXT_NET="$(echo net_${OS_USERNAME}-psd-extra)"
```
brianphaley, line 28: I don't think you need the echo here, can just use the variable directly and remove the ()

"${OS_USERNAME}-psd"
"${OS_USERNAME}-psd-extra"

brianphaley (changes requested):

openstack/profiles/prodstack6
sombrafam: ow, you're right we could just expand them

openstack/profiles/stsstack
```diff
@@ -4,6 +4,7 @@
 [[ -z "$NAMESERVER" ]] && export NAMESERVER="10.230.64.2"
 [[ -z "$SWIFT_IP" ]] && export SWIFT_IP="10.230.19.58"
 
-export UNDERCLOUD_EXT_NET="$(sed -E --quiet "s/.+OS_PROJECT_NAME=(.+)/\1_admin_net/p" ~/novarc)"
+source ~/novarc
```
brianphaley, line 7: Being pedantic but this should be at top of file like the others

brianphaley (approved):

PR #310, opened 2025-06-11
fix(octavia): handle --wait in configure_octavia based on Juju version

xtrusia:
Avoid passing --wait 10m for juju2 which doesn't support it.


nicolasbock (approved):
Thanks!

PR #312, opened 2025-06-25
Move user documentation to RTD

nicolasbock:


nicolasbock:
Thanks for the review @rodrigogansobarbieri . I addressed your comments. Please have another look. Thanks!

nicolasbock:
I think I addressed all of your comments @brianphaley . Very nice review, by the way :Smile:

docs/user-docs/openstack/octavia.md
```diff
@@ -0,0 +1,58 @@
...
+
+    ```console
+    tools/create_octavia_lb.sh __VM_UUID__
+    ```
+
+1. Now check that your load balancer is up and load balancing to port 80 in the VM
+
+    ```console
+    lb_vip=$(openstack loadbalancer show lb1 -c vip_address -f value)
+    lb_fip=$(openstack floating ip list --fixed-ip-address ${lb_vip} --format value --column "Floating IP Address")
```
brianphaley, line 56: nit: you can use -f and -c for format/column here, and i'm surprised floating_ip_address didn't work for the column, that might be a bug in the client :(

brianphaley (dismissed):

docs/user-docs/openstack/octavia.md
nicolasbock: Hmm, that actually does work.

PR #321, opened 2025-09-08
Map numerical OpenStack release to names

nicolasbock:


brianphaley (commented):
Seems useful just some nits

PR #326, opened 2025-09-26
Use most recent cirros image

nicolasbock:

### DIFF
--- a/openstack/profiles/default
+++ b/openstack/profiles/default
@@ -98,7 +98,7 @@ if $(juju show-application manila &>/dev/null); then
 fi
 upload_image cloudimages focal focal-server-cloudimg-amd64.img &
 upload_image cloudimages jammy jammy-server-cloudimg-amd64.img &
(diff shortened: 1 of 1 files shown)


dosaboy:
I see to remember that the reason why we have the version we use currently is because there were versions of openstack that didnt like the newer ones. See dcfced39d8d7d591f5db0ef09a4ffbee0c4442a0 and https://github.com/canonical/stsstack-bundles/issues/267

raineszm:
> I see to remember that the reason why we have the version we use currently is because there were versions of openstack that didnt like the newer ones. See [dcfced3](https://github.com/canonical/stsstack-bundles/commit/dcfced39d8d7d591f5db0ef09a4ffbee0c4442a0) and #267

I tried using a cirros-0.6.3 and it didn't seem to come up, so it seems like that's still an issue.

The issue with cirros-0.4.0 being that you can't ssh in without setting `-o PubkeyAcceptedAlgorithms=+ssh-rsa`

PR #329, opened 2025-10-07
juju-lnav: trim .priv when guessing ssh pub key

### LINKED ISSUE #328: juju-lnav fails to copy testkey.pub
The logic I added in #322 to guess the public key path from the private key path turned out to be a little to naive, i.e. it just added `.pub` to the end. 

```
checking ssh key for 10.149.138.27
Unable to ssh into root@10.149.138.27.
copying /home/ubuntu/testkey.priv.pub to octavia/0's authorized_keys
/home/ubuntu/testkey.priv.pub does not exist.
```

Other scripts create the pair `testkey.priv` and `testkey.pub` so a smarter guess would probably be to strip `.priv` from the end if it's there first and then add `.pub`. I'll put up a PR with that change.

raineszm:
The logic added in #322 to guess the public key path from the private key path turned out to be a little to naive, i.e. it just added .pub to the end. Other scripts create the pair testkey.priv and testkey.pub so a smarter guess would probably be to strip .priv from the end if it's there first and then add .pub.

Fixes #328


nicolasbock (approved):
Thanks!

PR #335, opened 2025-12-03
Relate cinder-volume to vault:certificates

lathiat:
When using --cinder-volume, we didn't link it to vault:certificates and
it failed to connect to keystone/nova etc due to an SSL error.

Add the relation when vault is in use.


dosaboy (approved):
lgtm but I am curious why this has not been a problem for anyone else? is it required for a particular cinder use case or has this just always been broken and we didnt notice?

PR #336, opened 2025-12-04
Fix watcher deployment

wilkmar:
This patch adds mandatory watcher config settings. Without those, the charm fails to deploy. See https://docs.openstack.org/charm-guide/latest/admin/compute/watcher.html#deployment for details.

This patch also adds a certificates relation to the vault.


dosaboy:
charm has an HA interface: https://charmhub.io/watcher/integrations#ha can we add support for that also?

PR #339, opened 2025-12-05
Add watcher-dashboard support

wilkmar:
Add watcher-dashboard support


dosaboy:
this needs placement info but can do that seperately

PR #377, opened 2026-04-22
Rename charm artifact inline when rename.sh is missing

xtrusia:
Some charm repos no longer ship a rename.sh helper(although it is in master (in progress), I think it's test suite's responsibility), so the build step leaves the artifact named ${charm}_${arch}.charm.
The test phase and bundle references expect the canonical ${charm}.charm. 
Do the rename inline after build (local or --remote-build) when rename.sh is absent, mirroring what rename.sh would have done. 

e.g https://review.opendev.org/c/openstack/charm-heat/+/985457


pponnuvel (approved):
LGTM

nicolasbock (approved):

PR #378, opened 2026-04-22
Don't default to charmcraft 1.5/stable

dosaboy:
If no charmcraft version info is found we were defaulting to 1.5/stable but this is annoying if you want to test a newer version of charmcraft. So now we do nothing if no version is found i.e. use whatever is installed.


wilkmar (approved):
LGTM,
I tested the commands/logic and it works as expected.

PR #379, opened 2026-04-23
Catch popd stack empty

dosaboy:
Classic charms don't have a src dir so popd stack will be empty. Catch popd stack empty error to allow script to continue


wilkmar (approved):
LGTM

PR #380, opened 2026-04-23
Only show test results if test phase was run

dosaboy:
In other works only show results if we do a full run or re-run the test phase.


nicolasbock (commented):
The change looks good, but the bashate linter is failing.

dosaboy:
not ready yet

nicolasbock (approved):

PR #381, opened 2026-04-23
Check that user specified arguments if required

brianphaley:
Some options in the test runner code require arguments, otherwise they throw an exception, for example:

$ ./openstack_regression_tests_runner.sh --func-test-target ./openstack_regression_tests_runner.sh: line 50: $2: unbound variable

If there is not at least one more argument remaining, error out with a message so the user can correct the call.

### DIFF
--- a/openstack/tools/charmed_openstack_functest_runner.sh
+++ b/openstack/tools/charmed_openstack_functest_runner.sh
@@ -63,6 +63,8 @@ OPTIONS:
         enough capacity to boot the vms required by the tests.
     --sleep TIME_SECS
         Specify amount of seconds to sleep between functest steps.
+    --debug
+        Run with debug (-x) enabled.
     --help
         This help message.
 EOF
@@ -143,6 +145,10 @@ while (($# > 0)); do
             set -x
             ;;
         --func-test-target)
+            if [[ $# < 2 ]]; then
+                echo "ERROR: '$1' requires argument"
+                exit 1
+            fi
             FUNC_TEST_TARGET+=( $2 )
             shift
             ;;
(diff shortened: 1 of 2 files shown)


openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -143,6 +145,10 @@ while (($# > 0)); do
             set -x
             ;;
         --func-test-target)
+            if [[ $# < 2 ]]; then
```
pponnuvel, line 148: Perhaps this could be wrapped in a function in `openstack/tools/func_test_tools/common.sh` to avoid repeating the same check, like:

```
require_opt_arg() {
    if [[ $# -lt 2 ]]; then
        echo "ERROR: '$1' requires argument"
        exit 1
    fi
}
```

And then called with: `require_opt_arg "$@"`

pponnuvel (approved):
LGTM

openstack/tools/charmed_openstack_functest_runner.sh
nicolasbock:
```suggestion
            if (( $# < 2 )); then
```

It's better to use an arithmetic expression if comparing this way.

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -157,11 +163,19 @@ while (($# > 0)); do
             WAIT_ON_DESTROY=false
             ;;
         --remote-build)
+            if [[ $# < 2 ]]; then
```
nicolasbock, line 166: Same here.

PR #384, opened 2026-04-29
Add prodstack7 profile and make it default

dosaboy:


brianphaley (dismissed):
lgtm, basically the same as the ps6 one

dosaboy:
> lgtm, basically the same as the ps6 one

yes although with this patch merged, in ps6 it will now be required to do ./configure prodstack6 (since prodstack7 is the default here)

dosaboy:
no longer modifies default

nicolasbock (approved):

PR #390, opened 2026-05-07
Add FIP network range to no_proxy

dosaboy:
This is needed in ps7

Closes: SET-2275

### DIFF
--- a/openstack/tools/charmed_openstack_functest_runner.sh
+++ b/openstack/tools/charmed_openstack_functest_runner.sh
@@ -216,6 +216,14 @@ FIP_MIN_ABC=${FIP_MIN%.*}
 FIP_MIN_D=${FIP_MIN##*.}
 FIP_MIN=${FIP_MIN_ABC}.$(($FIP_MIN_D + 64))
 
+if [[ -n $http_proxy ]]; then
+    if [[ -n $no_proxy ]]; then
+        export no_proxy=$no_proxy,$CIDR_EXT
+    else
+        export no_proxy=$CIDR_EXT
+    fi
+fi
+
 # Setup vips needed by zaza tests.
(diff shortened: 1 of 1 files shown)


pponnuvel (approved):

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -216,6 +216,14 @@ FIP_MIN_ABC=${FIP_MIN%.*}
 FIP_MIN_D=${FIP_MIN##*.}
 FIP_MIN=${FIP_MIN_ABC}.$(($FIP_MIN_D + 64))
 
+if [[ -n $http_proxy ]]; then
+    if [[ -n $no_proxy ]]; then
+        export no_proxy=$no_proxy,$FIP_MIN:$FIP_MAX/$CIDR_EXT
```
brianphaley: I'm a little confused based on what I think these values are set to, for example:

CIDR_EXT = 10.149.122.0/25
FIP_MIN = 10.149.122.64
FIP_MAX = 10.149.122.126

Which turns-into this:

10.149.122.64:10.149.122.126/10.149.122.0/25

Did we want just the /25 there? Or did I mis-understand the values here?

And as an aside, FIP_RANGE is defined below as $FIP_MIN:$FIP_MAX so could potentially use that if we move this.

brianphaley (changes requested):
-1 based on my question

openstack/tools/charmed_openstack_functest_runner.sh
dosaboy: fixed.

brianphaley (approved):

nicolasbock (approved):

PR #392, opened 2026-05-13
Fix --help and --list-overlays extra delay

wilkmar:
These switches (`--help`, `--list-overlays`) cause the `generate-bundle.sh` to exit early. 
If you just want to check help or list the supported overlays, and the juju controller is not reachable, before this patch, the command would take ages, waiting for the juju controller's response timeout.
I moved that part further down the execution flow once we know we don't exit early.


nicolasbock-canonical (approved):

pponnuvel (approved):

dosaboy (approved):
lgtm

PR #397, opened 2026-05-20
Remove charm functest build rename

dosaboy:
renaming is a legacy functionality that we are phasing out as we fix charms ci so we dont want to mandate it anymore. If a charm has incorrect build names in its test bundles then these should be fixed instead.


wilkmar (approved):
That's fine for me.

PR #398, opened 2026-05-21
Update cirros to the latest available (0.6.3)

wilkmar:
Updating the Cirros image to the latest available version.


wilkmar:
I tested the change on Yoga and Caracal. Cirros 0.6.3 boots without any issues and sets up ssh keys properly from the metadata service.  Logging in to a VM works with a password and using the private key.

dosaboy (approved):

PR #403, opened 2026-06-16
Set TMPDIR to a location juju can access

dosaboy:


openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -263,6 +263,10 @@ export TEST_JUJU3=1
 # NOTE: this should not be necessary for > juju 2.x but since we still have a need for it we add it in
 export TEST_ZAZA_BUG_LP1987332=1
 
+# Juju 3.x is a confined snap so cannot read/write from/to /tmp. This makes the tests use a path that juju can read/write.
+export TMPDIR=$HOME/tmp
```
nicolasbock-canonical: You could also

```suggestion
export TMPDIR=$(mktemp ${HOME}/bundles-XXXXXX)
```

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -263,6 +263,10 @@ export TEST_JUJU3=1
...
+mkdir -p $TMPDIR
```
nicolasbock-canonical: And then get rid of this line.

nicolasbock-canonical (approved):

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -433,3 +436,8 @@ if [[ -z $RERUN_PHASE ]] || [[ $RERUN_PHASE = 'test' ]]; then
 fi
 ) 2>&1 | tee $LOGFILE
 echo -e "\nResults also saved to $LOGFILE"
+
+if [ -d $TMPDIR ]; then
+    rm -rf $TMPDIR
+    unset TMPDIR
```
pponnuvel, line 442: This isn't needed.

pponnuvel (approved):
LGTM

openstack/tools/charmed_openstack_functest_runner.sh
dosaboy: it is if we dont want to have a new directory every time we run the script

pponnuvel: I was referring the `unset TMPDIR` line. At script exit, this doesn't make any difference.

PR #405, opened 2026-06-17
Make sure no_proxy supports all formats

dosaboy:
Some applications e.g. curl require network ranges to be exanded.


dosaboy:
the ci failure here exists on master and is fixed in https://github.com/canonical/stsstack-bundles/pull/404

wilkmar (approved):
LGTM

PR #406, opened 2026-06-18
Add support for uv based charm builds

wilkmar:
Zaza functional tests require Python 3.10, while some charm branches
require Python 3.12 to build. This means functional tests must run on
Jammy (Python 3.10), but the func runner would fail to build charms
that require Python 3.12.

To work around this, add support for building charms with uv, which
allows running the 'build' tox target with an arbitrary Python version.
The required Python version is detected from the charm repo's tox.ini
using 'tox --showconfig -e build', falling back to 3.12 on failure.

uv-based builds are currently not supported with '--remote-build'

### DIFF
--- a/openstack/tools/charmed_openstack_functest_runner.sh
+++ b/openstack/tools/charmed_openstack_functest_runner.sh
@@ -14,6 +14,7 @@ SKIP_BUILD=false
 SLEEP=
 WAIT_ON_DESTROY=true
 RERUN_PHASE=
+DEFAULT_BUILD_PYTHON_VERSION="3.12"
 
 . $(dirname $0)/func_test_tools/common.sh
 
@@ -212,6 +213,7 @@ done
 
 # Install dependencies
 which yq &>/dev/null || sudo snap install yq
+which uv &>/dev/null || sudo snap install astral-uv --classic
 
 # Ensure zosci-config checked out and up-to-date
 get_and_update_repo https://github.com/openstack-charmers/zosci-config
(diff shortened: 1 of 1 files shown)


pponnuvel (approved):
Looks good

brianphaley (commented):
lgtm, but i'll let others comment

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -291,6 +293,37 @@ juju_version=$(juju --version)
 LOGFILE=$(mktemp --suffix=-charm-func-test-results)
 (
 # 2. Build
+get_python_version_from_tox() {
+    local tox_ini="./tox.ini"
```
raineszm: Why define a tox_ini variable here? It doesn't apepar to be configurable or changed anywhere

openstack/tools/charmed_openstack_functest_runner.sh
```diff
@@ -291,6 +293,37 @@ juju_version=$(juju --version)
...
+
+    if [[ ! -f "$tox_ini" ]]; then
+        echo "ERROR: tox.ini not found; won't proceed" >&2
+        return 1
+    fi
+
+    _extract_basepython() {
```
raineszm: How about something like `tox config -e build -k basepython|grep -Po '(?<=python)[0-9.]+'` to determine the build version. This should automatically handle inheritance from testenv.

raineszm (commented):
Thanks for this. I've definitely run into issues with python version incompatibility when trying to use the functest runner script.

openstack/tools/charmed_openstack_functest_runner.sh
wilkmar: Hi @raineszm. Thanks for the suggestion. I had to change the command a bit to support both tox 3.x and 4.x . Tested on both versions. The patch is much simpler and 'user-friendly' now.

raineszm (approved):
LGTM

brianphaley (approved):

PR #408, opened 2026-06-24
Run func-target via tox-uv to avoid install_layout failure

xtrusia:
The func-target test run installs zaza, charm-tools, netifaces and friends from source. Under the host's setuptools that wheel build fails with AttributeError: install_layout. Running the env through tox-uv installs those deps with uv and sidesteps it.


pponnuvel (approved):
LGTM

PR #410, opened 2026-06-25
Clean leftover gateway ext-ports before func tests

xtrusia:
zaza creates an undercloud neutron port <server>_ext-port and attaches it to each gateway instance with interface_attach.
nova does not delete such a user-supplied port on teardown - it only detaches it and leaves it behind.
juju usually cleans it up - its openstack provider deletes ports filtered by device_id == server
but if nova clears device_id before that query runs, or the port delete errors out, or teardown is forced, the detached port survives on the shared external network. 
A later test whose floating IP lands on that leaked address then collides with the dead port at L2/ARP and can't reach its guest, so the failure shows up intermittently.
Before each model is built, delete ext-ports that are detached or whose device_id points at a gone instance; a port still attached to a live server is left alone.

This does not fully fix the leak. The teardown is async, so it can miss the port from the last model.
But it cleans old leaked ports before each build, so the collision is much less likely.

### DIFF
--- a/openstack/tools/clean_orphan_dataports.sh
+++ b/openstack/tools/clean_orphan_dataports.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+# Delete leaked gateway ext-ports on the undercloud before a func test.
+#
+# Usage: clean_orphan_dataports.sh [OPENRC]   (default: ~/novarc)
+set -eu
+
+OPENRC=${1:-$HOME/novarc}
+if [[ ! -r $OPENRC ]]; then
+    echo "WARNING: no undercloud openrc ($OPENRC) - skipping ext-port cleanup" >&2
+    exit 0
+fi
+source "$OPENRC"
+if ! command -v openstack &>/dev/null; then
+    echo "WARNING: no openstack CLI - skipping ext-port cleanup" >&2
+    exit 0
+fi
+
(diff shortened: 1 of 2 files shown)


openstack/tools/clean_orphan_dataports.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+# Delete leaked gateway ext-ports on the undercloud before a func test.
+#
+# Usage: clean_orphan_dataports.sh [OPENRC]   (default: $OPENRC, ~/novarc)
+set -eu
+
+OPENRC="${1:-${OPENRC:-$HOME/novarc}}"
```
dosaboy: i think this can be shortened to:

OPENRC=${1:-$HOME/novarc}

dosaboy (commented):
minor nit but otherwise ltgm

openstack/tools/clean_orphan_dataports.sh
pponnuvel: It's not equivalent though. It's $1 then existing OPENRC, and then $HOME/novarc. So if `OPENRC` could be set elsewhere (e.g. from parent shell),  then `OPENRC=${1:-$HOME/novarc}` would bypass that.

pponnuvel (approved):
LGTM

dosaboy (approved):

PR #412, opened 2026-06-29
Fix functest job identifier

dosaboy:
We were getting jobs from list of "job:" items but the list of jobs to run is actually project.check.jobs and the individual job items are overrides with extra information which some charms do not provide. So we now use project.check.jobs as the authoritative list of jobs to run.


wilkmar (approved):
LGTM

PR #418, opened 2026-07-03
Functest phase retry allow retry unrecognised

dosaboy:
When asked which phase to retry if the user provides an unrecognised phase name, allow them to provide it again.


wilkmar (approved):
LGTM, thx

